Issue #2, opened 2023-02-04
Rename packages and references to modid


Matthias1590:
Not sure what to though but it might be nice to do this early so we don't have to rename as much

RUGMJ:
I think the redstoneqol name is probably the best to go with internally and can easily  be changed if need be.

RUGMJ:
or maybe shorter to make it easier to type, "rqol" maybe?

Matthias1590:
it doesn't really need to be short internally, "redstonetools" seems fine

ItsEntDev:
commit done and PR finished

PR #8, opened 2023-02-04
Update fabric.mod.json

gdoge1:
Adding some info to it + link to mattbatwings's channel

### DIFF
--- a/src/main/resources/fabric.mod.json
+++ b/src/main/resources/fabric.mod.json
@@ -1,16 +1,15 @@
 {
   "schemaVersion": 1,
-  "id": "modid",
+  "id": "RedstoneQoL",
   "version": "${version}",
 
-  "name": "Example Mod",
-  "description": "This is an example description! Tell everyone what your mod is about!",
+  "name": "RedstoneQoL",
+  "description": "A Redstone Quality of Life mod.",
   "authors": [
-    "Me!"
+    ""
   ],
   "contact": {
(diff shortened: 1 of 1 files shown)


src/main/resources/fabric.mod.json
```diff
@@ -1,16 +1,15 @@
 {
   "schemaVersion": 1,
-  "id": "modid",
+  "id": "RedstoneQoL",
```
Matthias1590, line 3: modid's need to be lowercased, also not sure if we want to go with the redstoneqol name

src/main/resources/fabric.mod.json
```diff
@@ -1,16 +1,15 @@
...
+  "name": "RedstoneQoL",
+  "description": "A Redstone Quality of Life mod.",
   "authors": [
-    "Me!"
+    ""
   ],
   "contact": {
-    "homepage": "https://fabricmc.net/",
-    "sources": "https://github.com/FabricMC/fabric-example-mod"
+    "homepage": "https://www.youtube.com/channel/UCz28y3fThmxHaIh7Zl-i6KA",
```
Matthias1590, line 12: might be nice to add the source here

Matthias1590 (changes requested):

src/main/resources/fabric.mod.json
Matthias1590: I think we can keep things the way they are atm and rename everything at once (#2 )

PR #15, opened 2023-02-06
Added Authors and title.

M4ximumPizza:
I added two authors but can add more if there are more that you guys would like to add. And for the title, the majority of the votes were for the RsTools, but can be changed if wanted to by the authors.


Matthias1590:
I think we should figure out who we credit after the first release since there will definitely be more contributors by then

PR #19, opened 2023-02-06
Changed all modid references to 'redstonetools'

ItsEntDev:
Also, I changed the display name to Redstone Tools, and added two authors
"mattbatwings" and "admins: add more please"


Matthias1590:
Package name has a different order, we haven't decided on a domain yet but the project is called redstonetools so com.domain.redstonetools would work better

ItsEntDev:
Alright, will do quickly. Will probably be my last commit for today

Matthias1590:
Also I don't think we should add authors yet as we won't know who contributed until we release

Matthias1590 (changes requested):
Check comments

ItsEntDev:
Final commit has been made, merge if ok.
gn.

Matthias1590:
Thanks, good night!

src/main/resources/fabric.mod.json
```diff
@@ -1,30 +1,30 @@
...
+  "name": "Redstone Tools",
+  "description": "A Redstone Quality of Life mod. To improve efficiency and productivity.",
   "authors": [
-    "Me!"
+    "none"
   ],
   "contact": {
-    "homepage": "https://fabricmc.net/",
-    "sources": "https://github.com/FabricMC/fabric-example-mod",
+    "homepage": "https://example.com",
```
Matthias1590, line 12: I think we can just remove this key and the authors one

src/main/java/com/domain/redstonetools/example/ExampleMod.java
```diff
@@ -1,4 +1,4 @@
-package net.fabricmc.example;
+package com.domain.redstonetools.example;
```
Matthias1590, line 1: should be com.domain.redstonetools (net.fabricmc.example -> com.domain.redstonetools)

Matthias1590 (changes requested):

src/main/resources/fabric.mod.json
ItsEntDev: Noticed you left a comment, I'm not sure that removing keys is a good idea?

src/main/java/com/domain/redstonetools/example/ExampleMod.java
ItsEntDev: i'm tired, must have forgotten

ItsEntDev: - for me the packages are already correct, but I'm on my forked dev branch. are you looking at the wrong branch?

src/main/resources/fabric.mod.json
Matthias1590: I don't see the use in linking to a non relevant page though, we can always add it back later if we do end up getting a domain

RUGMJ: All of those should link to this GitHub repo

Matthias1590: > All of those should link to this GitHub repo

There's a "sources" key for that, we don't really need the homepage

Matthias1590 (dismissed):

Matthias1590 (approved):

PR #24, opened 2023-02-07
fixed mixins error

RUGMJ:


Matthias1590:
We don't have any mixins

RUGMJ:
no but there was an error being caused by the mixins setup on my machine, changing this fixed it so it might fix it for others

PR #28, opened 2023-02-09
logo changed to use the new version by orbyfied#8316

ItsEntDev:


RUGMJ:
The uri for the image should be in this repo and not your personal one

ItsEntDev:
> The uri for the image should be in this repo and not your personal one

I can't, because the image hasn't been pushed to this repo

RUGMJ:
But you can still link it? Relative linking instead of absolute would work just fine

README.md
```diff
@@ -1,4 +1,4 @@
-![QMD_logo_final_final2](https://user-images.githubusercontent.com/47035787/216833383-159ceea7-ab53-4197-8b5c-1835f1d28491.png)
+![logo](https://raw.githubusercontent.com/ItsEntDev/RSTools/dev/rst_logo.png)
```
Matthias1590, line 1: Please use a relative path to the image in the repo

Matthias1590 (changes requested):

README.md
ItsEntDev: done, even though I don't really see why it matters.

RUGMJ: > done, even though I don't really see why it matters.

It matters because if your account were ever hacked or if you for whatever reason wanted to, you would have the ability to change the logo to whatever you wanted to

RUGMJ: and it also makes much more sense for the image to be in the actual repo

Matthias1590 (approved):

PR #31, opened 2023-02-12
Compacted the code

M4ximumPizza:
I just moved some unnecessary code and compacted it


Matthias1590 (commented):
The reason I didn't compact it as much as possible is cause it is more readable this way (imo)

PR #36, opened 2023-02-12
Added BaseConvert Command

RUGMJ:


Matthias1590 (changes requested):
Add ArgumentType implementations for hexadecimal, decimal, binary and octal numbers, then add a command that takes a number and an optional output base that outputs the number to either the specified output base or all 4 bases (use an enum and an enum ArgumentType for this)

RUGMJ:
How would it know what the input base is?

RUGMJ:
The prefixes java use could be could, none for decimal, 0x for hex 0 for octal and 0b for binary

Matthias1590 (approved):

Issue #34, opened 2023-02-12
Get glass/wool variant of block you're looking at (https://discord.com/channels/1071119659346169907/1071203190147653856)


Matthias1590:
Get the glass variant of the block you're looking at if it's a wool block and get the wool variant if it's a glass block (red wool -> red stained glass and vice versa). Should be a command (/glass) without any arguments that casts a ray (no further than the player's reach) and behaves like Minecraft's middle click.

MiranCZ:
I can do that

Matthias1590:
Closed by #37

PR #37, opened 2023-02-13
Added /glass command

MiranCZ:
Added /glass command #34. Also added Utility for getting item by its name and for color codes becouse I needed them in my code.


src/main/java/com/domain/redstonetools/features/commands/glass/GlassFeature.java
```diff
@@ -0,0 +1,86 @@
...
+
+
+    //similar to: net.minecraft.client.MinecraftClient/doItemPick()
+    @Override
+    protected int execute(ServerCommandSource source, EmptyOptions options) {
+        MinecraftClient client = MinecraftClient.getInstance();
+        if (client.player == null) return -1;
+        String redPrefix = COLOR_CODE_CHAR + "" + ColorCodeUtils.Color.RED.code;
+        if (!client.player.getAbilities().creativeMode) {
+            client.player.sendMessage(Text.of(redPrefix + "You need to be in creative mode to execute this command!"), false);
```
Matthias1590, line 36: This isn't really necessary, it's not a vanilla mod so we don't really care if we give survival players free glass and wool

src/main/java/com/domain/redstonetools/RedstoneToolsClient.java
```diff
@@ -20,12 +20,13 @@ public class RedstoneToolsClient implements ClientModInitializer {
...
     // Feature annotation, it might also be useful for other reflection related tasks
     public static final List<Class<? extends AbstractFeature<?>>> FEATURE_CLASSES = List.of(
-        QuickTpFeature.class
+        QuickTpFeature.class, GlassFeature.class
     );
 
     @Override
     public void onInitializeClient() {
         LOGGER.info("Initializing Redstone Tools");
+        ItemUtils.register();
```
Matthias1590, line 29: Don't have a register method for ItemUtils as it introduces temporal coupling and is not necessary

src/main/java/com/domain/redstonetools/utils/ColorCodeUtils.java
```diff
@@ -0,0 +1,37 @@
+package com.domain.redstonetools.utils;
```
Matthias1590, line 1: I don't think we need this class, its nice but it'll make it harder to stay consistent in reporting errors, plain text or throwing command syntax exceptions works fine

src/main/java/com/domain/redstonetools/utils/ItemUtils.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.HashMap;
+
+public class ItemUtils {
+
+    private static final HashMap<String, Integer> ITEM_MAP = new HashMap<>();
+
+    private ItemUtils(){
+    }
+
+    public static void register() {
```
Matthias1590, line 15: Remove this method, make other methods search the registry instead of a map

Matthias1590 (changes requested):

src/main/java/com/domain/redstonetools/utils/ItemUtils.java
MiranCZ: I thought it would be better to have one method than searching trough it every time. Register method instead of straight creation is becouse I was not sure if the array is loaded at that time. I will remove the ColorUtils class, and make the text just white

src/main/java/com/domain/redstonetools/utils/ColorCodeUtils.java
MiranCZ: I will remove the ColorUtils class, and make the text just white

src/main/java/com/domain/redstonetools/features/commands/glass/GlassFeature.java
MiranCZ: ok

src/main/java/com/domain/redstonetools/utils/ItemUtils.java
Matthias1590: I guess we could have a method that finds an item by name but it should return that item instead of an arbitrary integer

MiranCZ: So the method I have, but its map with Items not integers?

Matthias1590: Yes, you could still look up items by their names but it'll return an item instead of an int

Matthias1590 (approved):

Matthias1590:
Please merge into dev

MiranCZ:
I cant. "Only those with write access to this repository can merge pull requests."

Matthias1590:
Changing the destination branch is what I meant

Issue #40, opened 2023-02-14
Entity Tile Drops


MiranCZ:
https://discord.com/channels/1071119659346169907/1071549143408975902

> New gamerule to toggle entity tile drops (Barrels, chests, etc).

MiranCZ:
Should it be added to gamerules or custom command?

Matthias1590:
Closed by #50

PR #50, opened 2023-02-16
Added gamerule for container drops

MiranCZ:
Added gamerule to toggle container drops - #40. I used mixin on `spawn` method in `ItemScatterer` which seemed to only affect containers.

### DIFF
--- a/src/main/java/com/domain/redstonetools/RedstoneToolsClient.java
+++ b/src/main/java/com/domain/redstonetools/RedstoneToolsClient.java
@@ -6,18 +6,15 @@ import com.domain.redstonetools.features.commands.quicktp.QuickTpFeature;
 import com.domain.redstonetools.utils.ReflectionUtils;
 import net.fabricmc.api.ClientModInitializer;
 import net.fabricmc.fabric.api.command.v1.CommandRegistrationCallback;
-import net.fabricmc.fabric.api.gamerule.v1.GameRuleFactory;
-import net.fabricmc.fabric.api.gamerule.v1.GameRuleRegistry;
-import net.minecraft.world.GameRules;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.List;
 
+import static com.domain.redstonetools.RedstoneToolsGameRules.DO_CONTAINER_DROPS;
+
 public class RedstoneToolsClient implements ClientModInitializer {
     public static final String MOD_ID = "redstonetools";
-    public static final GameRules.Key<GameRules.BooleanRule> DO_CONTAINER_DROPS =
-            GameRuleRegistry.register("doContainerDrops", GameRules.Category.DROPS, GameRuleFactory.createBooleanRule(true));
 
     public static final Logger LOGGER = LoggerFactory.getLogger(MOD_ID);
 
@@ -33,6 +30,8 @@ public class RedstoneToolsClient implements ClientModInitializer {
     @Override
     public void onInitializeClient() {
         LOGGER.info("Initializing Redstone Tools");
+
+        RedstoneToolsGameRules.register();
 
         CommandRegistrationCallback.EVENT.register(((dispatcher, dedicated) -> {
             for (var featureClass : FEATURE_CLASSES) {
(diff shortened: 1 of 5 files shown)


src/main/java/com/domain/redstonetools/mixin/ItemScatterMixin.java
```diff
@@ -0,0 +1,51 @@
...
+    @Shadow @Final private static Random RANDOM;
+
+
+    //there were problems with injecting into static method, so I just overwrote it
+
+    @Overwrite
+    public static void spawn(World world, double x, double y, double z, ItemStack stack) {
+        if (!world.getGameRules().getBoolean(DO_CONTAINER_DROPS)) return;
+
+        double d = EntityType.ITEM.getWidth();
```
Matthias1590, line 35: Run the original method instead of copying what it does

src/main/java/com/domain/redstonetools/RedstoneToolsClient.java
```diff
@@ -3,17 +3,21 @@
...
+import net.fabricmc.fabric.api.gamerule.v1.GameRuleRegistry;
+import net.minecraft.world.GameRules;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.List;
 
 public class RedstoneToolsClient implements ClientModInitializer {
     public static final String MOD_ID = "redstonetools";
+    public static final GameRules.Key<GameRules.BooleanRule> DO_CONTAINER_DROPS =
```
Matthias1590, line 19: Please put gamerules in a separate class (RedstoneToolsGameRules)

Matthias1590 (changes requested):
Move game rules to separate class and call original method

src/main/java/com/domain/redstonetools/mixin/ItemScatterMixin.java
orbyfied: That won't work by just calling it as it's overwritten though.
So we either have to look for a way to call the original method or attempt to fix the injection.

src/main/java/com/domain/redstonetools/RedstoneToolsClient.java
MiranCZ: Ok, I thought its not needed for just one gamerule

Matthias1590: It's fine for one gamerule but as we'll likely add more later it might be a good idea to just organize early before it gets out of hand

src/main/java/com/domain/redstonetools/mixin/ItemScatterMixin.java
```diff
@@ -0,0 +1,51 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import java.util.Random;
+
+import static com.domain.redstonetools.RedstoneToolsClient.DO_CONTAINER_DROPS;
+
+@org.spongepowered.asm.mixin.Mixin(ItemScatterer.class)
+public class ItemScatterMixin {
```
Matthias1590, line 24: Should be named `ItemScattererMixin`

src/main/java/com/domain/redstonetools/mixin/ItemScatterMixin.java
Matthias1590: Oh just noticed it isn't injecting, we should definitely inject instead of overwriting

MiranCZ: You can try, it was throwing exceptions for me becouse the method is static

src/main/java/com/domain/redstonetools/mixin/ItemScatterMixin.java
MiranCZ: Oh, lol

Matthias1590:
Are you sure this is only used for containers? I feel like this could very well be used for dying entities such as mobs as well

MiranCZ:
Yeah, I tried it and it seemed to work fine. (I tried it with /kill and piglin with armor)

MiranCZ:
the `RedstoneToolsGameRules` has a register method becouse otherwise the gamerule load upon call to the variable

Matthias1590 (changes requested):
Looks good, I'll merge once the mixin injects instead of overwriting

Matthias1590 (dismissed):

PR #51, opened 2023-02-16
Migrate to Version Catalog and update dependencies

BennyDioxide:
Version Catalog can make managing dependencies and their versions easily and make build.gradle more readable.
Also changed the domain name.


Matthias1590 (approved):

BennyDioxide:
Change the versions and dependencies in gradle/libs.versions.toml from now on

Issue #43, opened 2023-02-15
Make /base command report invalid number error


Matthias1590:
Currently when you provide an invalid number (`/base 10 _2j5 2`, _2j5 is not a valid base 10 number) it'll display a weird error message, it should throw a CommandSyntaxException instead

RUGMJ:
Should be as easy as wrapping everything in a try catch and throwing the exception in the catch block

Matthias1590:
Should be, yeah, do you want to be assigned? @RUGMJ

Matthias1590:
Closed by #58

PR #58, opened 2023-02-16
Catched number format exception

RUGMJ:
Solves #43


Matthias1590 (changes requested):
Try to only encapsulate the code that can fail in the try block, also what's the todo for?

RUGMJ:
I don't think it really makes a difference, the todo was left there by mistake from my editor I'll push a commit to remove it now

Matthias1590:
> I don't think it really makes a difference, the todo was left there by mistake from my editor I'll push a commit to remove it now

It does, if the sendFeedback method were to throw an exception that happened to be caught it would fail but for the wrong reason, try blocks should be as small as possible

RUGMJ:
It's only catching number format errors so it wouldn't make a difference

Matthias1590:
It's still considered bad practice, keeping try blocks small keeps the code readable

Matthias1590 (approved):

PR #65, opened 2023-02-18
Refactored RedstonerFeature

### LINKED ISSUE #54: Make the /redstoner command modify the world
Currently it just runs commands on the world as the source, we should modify the source's world directly instead. This also means we can remove the commandsource utils class

RUGMJ:
Closes #54


Matthias1590:
This doesn't set the time

Matthias1590 (changes requested):
I think you forgot the time, rest looks good

src/main/java/com/domain/redstonetools/features/commands/redstoner/RedstonerFeature.java
```diff
@@ -23,7 +23,7 @@ protected int execute(ServerCommandSource source, EmptyOptions options) {
         rules.get(GameRules.DO_DAYLIGHT_CYCLE).set(false, server);
         rules.get(GameRules.DO_MOB_SPAWNING).set(false, server);
         rules.get(RedstoneToolsGameRules.DO_CONTAINER_DROPS).set(false, server);
-
+        source.getWorld().setTimeOfDay(18000); // Midnight
```
Matthias1590, line 26: Should be noon (6000)

Matthias1590 (changes requested):

PR #66, opened 2023-02-18
Core: Refactor v1

orbyfied:
Currently:
- Create feature instances separately instead of in the command registration
- Minimize reflection and maximize caching


Matthias1590:
I'm already reworking the reflection and feature instantiation

PR #68, opened 2023-02-18
Removed unused imports

RUGMJ:


Matthias1590:
The auto formatter doesnt exactly make the code any more readable by putting newlines in random places

PR #72, opened 2023-02-19
Added //update command and automatic use after WorldEdit edit

### LINKED ISSUE #60: Update redstone after //paste
Should update all redstone in the WE selection, send any feedback messages using the WE api

MiranCZ:
resolved #60


Matthias1590:
I don't think we need the feature class, if we do we'll have to rewrite it to use the new core system

MiranCZ:
I think having just the command is also useful. Like when you forget to turn the gamerule on

Matthias1590:
True, are you ok with rewriting it? or should I? You can take a look at other features to see what changed, you basically have the arguments in the class now instead of a separate options class

MiranCZ:
Yeah, I think it will be fine since this doesnt have any arguments

MiranCZ:
It should be OK now

Matthias1590 (approved):

PR #77, opened 2023-02-20
Readability Changes

RUGMJ:
After the 38245b7 commit some imports were failing because the classes no longer existed, as well as this some imports were not needed finally some parts of the code were not indented properly


Matthias1590:
The autoformatting makes the code less readable imo, I think it's fine the way it is

RUGMJ:
Well the imports are still broken

Matthias1590:
Oh my bad, missed that part, feel free to create a pr for that, I just don't think the autoformatting helps much

PR #81, opened 2023-02-22
Added MinSelection Feature

### LINKED ISSUE #55: Minimize selection //minsel
https://discord.com/channels/1071119659346169907/1071203692847566868

RUGMJ:
Closes #55


src/main/java/com/domain/redstonetools/features/commands/MinSelectionFeature.java
```diff
@@ -0,0 +1,123 @@
...
+
+        selector.learnChanges();
+        selector.explainRegionAdjust(actor, localSession);
+
+        actor.printInfo(TextComponent.of("Done."));
+
+        return Command.SINGLE_SUCCESS;
+    }
+
+    private List<BlockVector3> makeChanges(World selectionWorld, Region selection)
```
Matthias1590, line 60: Why does this return a list of vectors? Also might be nice to give this a more descriptive name

src/main/java/com/domain/redstonetools/features/commands/MinSelectionFeature.java
```diff
@@ -0,0 +1,123 @@
...
+
+            finished = false;
+
+        }
+
+        try {
+            selection.contract(changes.toArray(new BlockVector3[changes.size()]));
+        } catch (RegionOperationException e) {
+            var msg = new LiteralMessage("There was an error modifying the region.");
+            throw new CommandSyntaxException(new SimpleCommandExceptionType(msg), msg);
```
Matthias1590, line 93: No need to pass the type, just pass null like all other features do

src/main/java/com/domain/redstonetools/features/commands/MinSelectionFeature.java
```diff
@@ -0,0 +1,123 @@
...
+
+import net.minecraft.server.command.ServerCommandSource;
+import net.minecraft.text.Text;
+
+@Feature(command = "/minsel", description = "Removes all air-only layers from a selection", name = "Minimize Selection")
+public class MinSelectionFeature extends CommandFeature {
+
+    @Override
+    protected int execute(ServerCommandSource source) throws CommandSyntaxException {
+        var actor = FabricAdapter.adaptPlayer(source.getPlayer());
```
Matthias1590, line 29: We should probably extract the functionality to get the WE selection into some utility method and make this feature and other WE related features use that instead

Matthias1590 (changes requested):

src/main/java/com/domain/redstonetools/features/commands/MinSelectionFeature.java
RUGMJ: If you read the function you'll see that it uses it to modify the changes list

src/main/java/com/domain/redstonetools/features/commands/MinSelectionFeature.java
RUGMJ: Wouldn't that give a blank error message?

src/main/java/com/domain/redstonetools/features/commands/MinSelectionFeature.java
RUGMJ: I've suggested this but I think it's out of the scope of this PR

src/main/java/com/domain/redstonetools/features/commands/MinSelectionFeature.java
Matthias1590: Nope, it works fine for other features

Matthias1590: This probably shouldn't throw a command syntax exception though, as it's not related to the user's input, a runtime exception might be better

src/main/java/com/domain/redstonetools/features/commands/MinSelectionFeature.java
Matthias1590: Fair

src/main/java/com/domain/redstonetools/features/commands/MinSelectionFeature.java
RUGMJ: > Nope, it works fine for other features

I mean to the player

RUGMJ: > This probably shouldn't throw a command syntax exception though, as it's not related to the user's input, a runtime exception might be better

I thought that too but the error is related to the selection so it kind of is the input, up to you tho

src/main/java/com/domain/redstonetools/features/commands/MinSelectionFeature.java
RUGMJ: > I can see that it makes changes but those changes aren't used anywhere, something like "minimizeRegion" or "removeAirLayers" makes more sense to me than "makeChanges"

ofc its used? `selection.contract(changes.toArray(new BlockVector3[changes.size()]));`

RUGMJ: renaming the function makes sense tho

src/main/java/com/domain/redstonetools/features/commands/MinSelectionFeature.java
RUGMJ: > Fair, though generally it's good practice to abstract something away if you find yourself copy pasting it

I completely agree but this PR is for adding the minimise selection feature so modifying other features wouldn't make sense, id be happy to work on this in a separate PR after this PR is resolved

src/main/java/com/domain/redstonetools/features/commands/MinSelectionFeature.java
Matthias1590: I think since it's not related to command syntax a runtime exception might be nicer here

src/main/java/com/domain/redstonetools/features/commands/MinSelectionFeature.java
Matthias1590: > > I can see that it makes changes but those changes aren't used anywhere, something like "minimizeRegion" or "removeAirLayers" makes more sense to me than "makeChanges"
> 
> ofc its used? `selection.contract(changes.toArray(new BlockVector3[changes.size()]));`

The value you return isn't used

RUGMJ: > The value you return isn't used
```
    if (!finished)
            changes = makeChanges(selectionWorld, selection);
```

RUGMJ: It isn't used by the execute function but it is used in the makeChanges method itself ig it could be turned into 2 methods but it works just fine as is

Matthias1590: it's assigned but never used there and it's never even assigned outside of the function

RUGMJ: > it's assigned but never used there and it's never even assigned outside of the function

The function will try to minimise the selection, if it couldn't then it is deemed finished so it returns the changes it was given, if not it will call itself with the changes it was given this loops until the selection is fully minimised, there could be a count added which if higher than x the function immediately returns null but after extensive testing this is not needed

Matthias1590: I'm not questioning the implementation, I'm saying that the return value isn't used and so the method may as well return void, this would keep the signature cleaner and the code simpler

RUGMJ: > I'm not questioning the implementation, I'm saying that the return value isn't used and so the method may as well return void, this would keep the signature cleaner and the code simpler

But the method calls itself and assigns a variable so returning void would break it

RUGMJ: so doing what youre asking would require the implementation changing

Matthias1590: > > I'm not questioning the implementation, I'm saying that the return value isn't used and so the method may as well return void, this would keep the signature cleaner and the code simpler
> 
> But the method calls itself and assigns a variable so returning void would break it

then don't assign a variable to it, the value isn't used anyway

RUGMJ: but the value is used, its used to actually modify the selection

Matthias1590: > but the value is used, its used to actually modify the selection

it's not, when you reassign changes you're not reassigning selection

DiamondTaco: matthias the variable is assigned and returned in the function bc its recursive

Matthias1590: > matthias the variable is assigned and returned in the function bc its recursive

the value isn't used outside of the method so I don't see why it should return anything at all

RUGMJ: > > but the value is used, its used to actually modify the selection
> 
> it's not, when you reassign changes you're not reassigning selection

but it is here `selection.contract(changes.toArray(new BlockVector3[changes.size()]));`

RUGMJ: > > matthias the variable is assigned and returned in the function bc its recursive
> 
> the value isn't used outside of the method so I don't see why it should return anything at all

Which is why I suggested a 2nd method which is void and wraps the other one tho this just makes things more confusing

Matthias1590: > > > but the value is used, its used to actually modify the selection
> > 
> > it's not, when you reassign changes you're not reassigning selection
> 
> but it is here `selection.contract(changes.toArray(new BlockVector3[changes.size()]));`

yes, I know the changes variable is used inside of the method but it isn't used after the last assignment so both the last assignment and return statements can be removed

RUGMJ: But they cant because the function loops itself so that it can minimise it completely, although the return isnt used on the first instance of the function it is used on the rest

DiamondTaco: matthias, it's a recursive function. the return value is used within itself to shrink the selection.

RUGMJ: > matthias, it's a recursive function. the return value is used within itself to shrink the selection.

Thanks ive been trying to put it into concise words like that

Matthias1590: that's not how recursion works, you're not using the return value, try replacing the last assignment with just a call and remove the return statement, it'll work the exact same

src/main/java/com/domain/redstonetools/features/commands/MinSelectionFeature.java
RUGMJ: > I think since it's not related to command syntax a runtime exception might be nicer here

Shouldnt it be a CommandException like ive changed it to? I thought thats what you meant? Or do you mean the java runtime exception

Matthias1590: CommandException works too I guess

Matthias1590 (approved):

PR #82, opened 2023-02-23
Dustplacer Feature

Ninty9:
Basic functionality for the feature is done

### DIFF
--- a/src/main/java/com/domain/redstonetools/features/commands/DustPlacerFeature.java
+++ b/src/main/java/com/domain/redstonetools/features/commands/DustPlacerFeature.java
@@ -20,50 +20,15 @@ public class DustPlacerFeature extends CommandFeature {
     static public Boolean isActive = false;
     static public List<Identifier> blockList = new ArrayList<>();
 
-    public boolean blockInBlockList(Identifier block) {
-        return false;
-    }
 
     @Override
     protected int execute(ServerCommandSource source) throws CommandSyntaxException {
-        if(blockList.isEmpty()) {
-            blockList.add(Blocks.WHITE_WOOL.getLootTableId());
-            blockList.add(Blocks.ORANGE_WOOL.getLootTableId());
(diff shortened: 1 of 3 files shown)


src/main/java/com/domain/redstonetools/features/commands/DustPlacerFeature.java
```diff
@@ -0,0 +1,76 @@
...
+    static public Boolean isActive = false;
+    static public List<Identifier> blockList = new ArrayList<>();
+
+    public boolean blockInBlockList(Identifier block) {
+        return false;
+    }
+
+    @Override
+    protected int execute(ServerCommandSource source) throws CommandSyntaxException {
+        if(blockList.isEmpty()) {
```
MiranCZ, line 29: maybe add a `static{` statement to add all those values into list

Ninty9: Good idea, added. (also cleaned up a method that was lying around for some reason, don't know why i overlooked that.)

Matthias1590 (changes requested):
The list of blocks should be stored in a settings file, the toggling of the feature should be done using 2 commands: /autodust enable and /autodust disable (override the registerCommands method to register them)

Matthias1590:
any updates on this?

orbyfied:
> The list of blocks should be stored in a settings file (use dependency injection to inject something like a config provider or config reader into the autodust feature), the toggling of the feature should be done using 2 commands: /autodust enable and /autodust disable (override the registerCommands method to register them)

I think we should first get a configuration system working.

Matthias1590:
> > The list of blocks should be stored in a settings file (use dependency injection to inject something like a config provider or config reader into the autodust feature), the toggling of the feature should be done using 2 commands: /autodust enable and /autodust disable (override the registerCommands method to register them)
> 
> I think we should first get a configuration system working.

that would be the configuration system

MiranCZ:
Why was this closed?

PR #84, opened 2023-02-24
Feature Airplace: Added block outline

orbyfied:
It now renders a block outline where it will place the block when the feature is enabled, and you're aiming at the air.


Matthias1590:
Why was this closed?

PR #86, opened 2023-02-26
fixed dependecies

RUGMJ:
credit to @PatrickMSM from the fabric dev server for his help

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -32,6 +32,10 @@ dependencies {
 	// Reflections
 	include implementation ('org.reflections:reflections:0.9.12')
 
+	include 'org.javassist:javassist:3.26.0-GA'
+	include 'javax.inject:javax.inject:1'
+	include 'aopalliance:aopalliance:1.0'
+
 	// To change the versions see the gradle.properties file
 	minecraft "com.mojang:minecraft:${project.minecraft_version}"
 	mappings "net.fabricmc:yarn:${project.yarn_mappings}:v2"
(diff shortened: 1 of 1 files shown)


Matthias1590:
I saw you post about a call to MinecraftClient.getInstance failing, is that also fixed with this PR?

build.gradle
```diff
@@ -24,13 +24,17 @@ repositories {
 
 dependencies {
 	// jUnit 5
-	implementation 'org.junit.jupiter:junit-jupiter:5.8.1'
+	include implementation ('org.junit.jupiter:junit-jupiter:5.8.1')
```
Matthias1590, line 27: `include implementation` can be replaced with `api` iirc

Matthias1590 (changes requested):

build.gradle
RUGMJ: from what ik thats only for mods which are apis, these are regular maven libraries

RUGMJ:
> I saw you post about a call to MinecraftClient.getInstance failing, is that also fixed with this PR?

this fixes that yes

build.gradle
Matthias1590: "api" is an alias for "include implementation"

RUGMJ: I tried that but it results in the crash that was occurring before

Matthias1590: I guess this is fine for now though

Matthias1590: I probably mixed some things up then, good to know that that's fixed now though

Matthias1590 (approved):

PR #87, opened 2023-02-26
Revise blockread

RUGMJ:
Revised blockread to the following spec
  - change arguments to
  - //read spacing on_block to_base reverse_bits
  - in other words, swap on_block with to_base positionally
  - and make defualt to_base 10


RUGMJ:
oops accidentally was committing to main, i'll re open

PR #93, opened 2023-02-27
Revised redstoner

RUGMJ:
added feedback in chat as if the player executed the gamerules themselves

### DIFF
--- a/src/main/java/com/domain/redstonetools/features/commands/RedstonerFeature.java
+++ b/src/main/java/com/domain/redstonetools/features/commands/RedstonerFeature.java
@@ -1,36 +1,27 @@
 package com.domain.redstonetools.features.commands;
 
-import com.domain.redstonetools.RedstoneToolsGameRules;
 import com.domain.redstonetools.features.Feature;
+import com.domain.redstonetools.utils.CommandSourceUtils;
 import com.mojang.brigadier.Command;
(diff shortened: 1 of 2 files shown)


RUGMJ:
image

Matthias1590:
Please make an issue before creating a pr next time. For the feedback, maybe it'd be better if it ran the commands instead of modifying the server object (the old way ran the commands) so it automatically generates the correct output

Matthias1590 (changes requested):
Run commands instead of modifying world object

MiranCZ:
Can't this just undo #54

RUGMJ:
Surely this can be a revert of 823b5fc742f86b0d05cd11b80a2d2cac8e5b6bbb then?

Matthias1590:
> > Please make an issue before creating a pr next time.
> 
> It was already discussed in the dev meeting so I thought putting something on the issue tracker would be unnecessary 

Having everything in one place keeps it clear what needs to be done and let's us assign people to tasks

Matthias1590:
> Surely this can be a revert of 823b5fc742f86b0d05cd11b80a2d2cac8e5b6bbb then?

Probably, yes

RUGMJ:
Sorry it took so long for such a simple fix, should be back to more active commits

Matthias1590 (approved):

PR #95, opened 2023-02-27
Revise ColorCode

DiamondTaco:
I merged into the wrong branch at first


MiranCZ:
You can just change that

MiranCZ:
Yeah, when you click the same edit button, you can change the branch

MiranCZ:
Why do you want to access it outside the class? It should be `public static` to do that tho

DiamondTaco:
> Why do you want to access it outside the class? It should be `public static` to do that tho

:facepalm: yeah, you're right. Matthias asked for it specifically on the issue.

Matthias1590:
> > Why do you want to access it outside the class? It should be `public static` to do that tho
> 
> :facepalm: yeah, you're right. Matthias asked for it specifically on the issue.

moving them to a utils class might be cleaner

Matthias1590 (changes requested):
Move utility functions to a utils class and turn color into an enum

DiamondTaco:
> 

done

DiamondTaco:
> Move utility functions to a utils class and turn color into an enum

done

src/main/java/com/domain/redstonetools/utils/ColorUtils.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.HashMap;
+import java.util.regex.Pattern;
+
+public class ColorUtils {
+    private static final Pattern MATCH_TARGET_PATH_PATTERN = Pattern.compile(
+            "^minecraft:(\\w+?)_(wool|stained_glass|concrete_powder|concrete|glazed_terracotta|terracotta)$"
+    );
+
+    // memoize matched block-id's
+    public static final HashMap<String, Pair<String, String>> blockMap = new HashMap<>();
```
Matthias1590, line 14: This doesnt need to be public

src/main/java/com/domain/redstonetools/utils/ColorUtils.java
```diff
@@ -0,0 +1,68 @@
...
+
+    public static Pair<String, String> getColorFromBlockId(String blockId) {
```
Matthias1590, line 16: Why does this return a pair and not a color like the method name suggests?

src/main/java/com/domain/redstonetools/utils/ColorUtils.java
```diff
@@ -0,0 +1,68 @@
...
+        } else {
+            output = null;
+        }
+
+        blockMap.put(blockId, output);
+
+        return output;
+    }
+
+    public enum Colors {
```
Matthias1590, line 43: Maybe move this to the block color argument file and rename to "Color", also the blockColor argument should probably use this enum instead of strings

Matthias1590 (changes requested):

src/main/java/com/domain/redstonetools/utils/ColorUtils.java
DiamondTaco: I'll rename it

src/main/java/com/domain/redstonetools/utils/ColorUtils.java
DiamondTaco: done

src/main/java/com/domain/redstonetools/utils/ColorUtils.java
DiamondTaco: I think it should stay in ColorUtils, but i made BlockColorArgumentType use it

src/main/java/com/domain/redstonetools/features/arguments/BlockColorArgumentType.java
```diff
@@ -1,26 +1,14 @@
 package com.domain.redstonetools.features.arguments;
 
+import com.domain.redstonetools.utils.ColorUtils;
+
+import java.util.Arrays;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 public class BlockColorArgumentType extends SetArgumentType<String> {
```
Matthias1590, line 9: this might as well extend SetArgumentType<Color>

Matthias1590 (changes requested):

src/main/java/com/domain/redstonetools/features/arguments/BlockColorArgumentType.java
DiamondTaco: > 

wdym?

Matthias1590: The block color argument type should give you a color instead of a string

DiamondTaco: > The block color argument type should give you a color instead of a string



> The block color argument type should give you a color instead of a string

i see, the line of code you highlighted was throwing me off

DiamondTaco: i tried adding that and it breaks tab autocomplete, all the color names are capitalized

DiamondTaco:
Can this get merged?

src/main/java/com/domain/redstonetools/features/arguments/BlockColorArgumentType.java
Matthias1590: I don't see why it wouldn't be possible though

DiamondTaco:
> Once the block color arg type returns a color I'll merge

I can do it, but then all the names while running the command are capitalized, and autocomplete breaks. Not having tab autocomplete for this feature would be a serious downgrade.

Matthias1590:
> > Once the block color arg type returns a color I'll merge
> 
> I can do it, but then all the names while running the command are capitalized, and autocomplete breaks. Not having tab autocomplete for this feature would be a serious downgrade.

Why are they capitalized and can they not be capitalized? Sounds like a simple bug with a simple fix

DiamondTaco:
> > > Once the block color arg type returns a color I'll merge
> > 
> > 
> > I can do it, but then all the names while running the command are capitalized, and autocomplete breaks. Not having tab autocomplete for this feature would be a serious downgrade.
> 
> Why are they capitalized and can they be not capitalized? Sounds like a simple bug with a simple fix

The reason they are capitalized is the enum variants are capitalized, so brigadier picks them up as capitalized. The only way I know how to fix this would be to fully lowercase all of the color variants, which goes against constant naming conventions. IDK which option to choose tbh.

MiranCZ:
Can't you just `do.toString().toLowerCase()`?

DiamondTaco:
> Can't you just `do.toString().toLowerCase()`?

its handled by mojang's library, so i can't do anything about it.

DiamondTaco:
> > Can't you just `do.toString().toLowerCase()`?
> 
> its handled by mojang's library, so i can't do anything about it.

nvm im an idiot i can just override toString on the enum

DiamondTaco:
> Once the block color arg type returns a color I'll merge

done

src/main/java/com/domain/redstonetools/features/arguments/BlockColorArgumentType.java
DiamondTaco: done

src/main/java/com/domain/redstonetools/features/commands/ColorCodeFeature.java
```diff
@@ -12,88 +13,59 @@
...
+            .withDefault(null);
 
-        if (colorEndIndex == -1)
-            return null;
 
-        return blockId.substring(0, colorEndIndex);
-    }
-
-    private String getColorlessBlockId(BlockState state) {
+    private boolean shouldBeColored(World world, BlockVector3 pos, String onlyColor) {
```
Matthias1590, line 39: Why is onlyColor a string? I think it'd make more sense to use the Color enum everywhere as opposed to converting it to a string in some places

Matthias1590 (changes requested):

src/main/java/com/domain/redstonetools/features/commands/ColorCodeFeature.java
DiamondTaco: It is being compared against the matched block-id's which are strings. it doesn't really matter tbh

Matthias1590: they should be converted only when comparing, passing them around as strings is more error prone

DiamondTaco: There's no bugs in the code tho

Matthias1590: That doesn't mean much, code shouldn't be error prone, enums are perfect for things like this, there's no reason to pass around strings

DiamondTaco: pushed

DiamondTaco: > That doesn't mean much, code shouldn't be error prone, enums are perfect for things like this, there's no reason to pass around strings

srry i just got a little annoyed

Matthias1590: Thanks

DiamondTaco:
why closed?

MiranCZ:
> why closed?

Good question

PR #112, opened 2023-03-09
Revise baseconvert

RUGMJ:
Solves this by using a long value instead of an integer 
```
/base 10 2147483648 2 
returns "Input number does not match the specified base" (2147483648 = 2^31)
/base 12 4bb2308a8 10
Returns the same thing 4bb2308a8 is 2^31 in base 12
This should return a more descriptive error letting the user know the inputted value is too large
```

### DIFF
--- a/src/main/java/com/domain/redstonetools/features/commands/BaseConvertFeature.java
+++ b/src/main/java/com/domain/redstonetools/features/commands/BaseConvertFeature.java
@@ -21,14 +21,14 @@ public class BaseConvertFeature extends CommandFeature {
 
     @Override
     protected int execute(ServerCommandSource source) throws CommandSyntaxException {
-        int input;
+        long input;
         try {
-            input = Integer.parseInt(number.getValue(), fromBase.getValue());
+            input = Long.parseLong(number.getValue(), fromBase.getValue());
         } catch (NumberFormatException e) {
             throw new CommandSyntaxException(null, Text.of("Inputted number does not match the specified base"));
         }
 
-        var output = Integer.toString(input, toBase.getValue());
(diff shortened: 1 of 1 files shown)


Matthias1590:
Please merge into dev

src/main/java/com/domain/redstonetools/features/commands/BaseConvertFeature.java
```diff
@@ -21,14 +21,14 @@ public class BaseConvertFeature extends CommandFeature {
 
     @Override
     protected int execute(ServerCommandSource source) throws CommandSyntaxException {
-        int input;
+        long input;
         try {
-            input = Integer.parseInt(number.getValue(), fromBase.getValue());
+            input = Long.parseLong(number.getValue(), fromBase.getValue());
```
Matthias1590, line 26: Use biginteger

Matthias1590 (changes requested):

src/main/java/com/domain/redstonetools/features/commands/BaseConvertFeature.java
RUGMJ: theres no parser for bigints tho? How would it convert the base?

Matthias1590: `public BigInteger(String val, int radix)`

RUGMJ: My bad, never used bigInts before

Matthias1590 (approved):

PR #113, opened 2023-03-09
Implemented simple configuration system for now, partially closes #101

### LINKED ISSUE #101: A config manager
Some sort of config management system that let's features and mixins read from a config and has a gui to let the user configure options would be great, for now just reading the options will be enough though, I'll create a separate issue for editing the config with a gui

orbyfied:
- Saves to and loads from YAML
- One file per feature (`.minecraft/config/redstonetools/<id>.yml`)
- Uses the `TypeSerializer` for the values
- Stores options in order with the section path
- Automatically creates options for defaults for optional `Argument`s in a feature class
- Automatically applies the config system to each registered feature

- I also added some logging so it's easier to see where an error comes from


orbyfied (commented):
very good yes

M4ximumPizza (approved):

PR #119, opened 2023-03-11
toggle change

Ninty9:
kept the onEnable and onDisable functions for eventual feedback functionality in the future


src/main/java/com/domain/redstonetools/features/toggleable/ToggleableFeature.java
```diff
@@ -17,26 +17,17 @@ protected void registerCommands(CommandDispatcher<ServerCommandSource> dispatche
...
+                .executes(this::run));
     }
 
     public boolean isEnabled() {
         return enabled;
     }
 
-    private int enable(CommandContext<ServerCommandSource> context) throws CommandSyntaxException {
-        enabled = true;
+    private int run(CommandContext<ServerCommandSource> context) throws CommandSyntaxException {
```
Matthias1590, line 27: might be nice to rename this to toggle

Matthias1590:
might be nice to just send feedback in the run method so features don't have to do that themselves

Matthias1590 (changes requested):

Ninty9:
> might be nice to just send feedback in the run method so features don't have to do that themselves

that's definitely doable, but shouldn't we wait until we have the feedback util for that?

Matthias1590:
> > might be nice to just send feedback in the run method so features don't have to do that themselves
> 
> that's definitely doable, but shouldn't we wait until we have the feedback util for that?

Might be easier to just do it now so we can refactor it later and not risk forgetting about it

Ninty9:
That should be everything

Matthias1590 (approved):

PR #153, opened 2023-04-27
Adds update popup

RUGMJ:
Adds a popup after the title screen indicating to the user if theres an update available, it knows this from the github releases api, currently this system wont work for betas or if we add support to more than 1 mc version at a time

### DIFF
--- a/src/main/java/tools/redstone/redstonetools/gui/UpdatePopupScreen.java
+++ b/src/main/java/tools/redstone/redstonetools/gui/UpdatePopupScreen.java
@@ -3,7 +3,6 @@ package tools.redstone.redstonetools.gui;
 import net.minecraft.client.gui.screen.Screen;
 import net.minecraft.client.gui.widget.ButtonWidget;
 import net.minecraft.text.Text;
-import tools.redstone.redstonetools.RedstoneToolsClient;
 
 public class UpdatePopupScreen extends PopupScreen {
 
--- a/src/main/java/tools/redstone/redstonetools/gui/UpdatePopupScreen.java
+++ b/src/main/java/tools/redstone/redstonetools/gui/UpdatePopupScreen.java
@@ -18,7 +18,6 @@ public class UpdatePopupScreen extends PopupScreen {
             this.close();
         }));
         this.addDrawableChild(new ButtonWidget(this.width / 2 - 155, y, 150, 20, Text.of("Ignore"), (button) -> {
-            RedstoneToolsClient.LOGGER.info("Ignored Update");
             this.close();
         }));
     }
--- a/src/main/java/tools/redstone/redstonetools/gui/UpdatePopupScreen.java
+++ b/src/main/java/tools/redstone/redstonetools/gui/UpdatePopupScreen.java
@@ -1,0 +1,26 @@
+package tools.redstone.redstonetools.gui;
+
+import net.minecraft.client.gui.screen.Screen;
+import net.minecraft.client.gui.widget.ButtonWidget;
+import net.minecraft.text.Text;
+import tools.redstone.redstonetools.RedstoneToolsClient;
+
+public class UpdatePopupScreen extends PopupScreen {
+
+    public UpdatePopupScreen(Screen parent) {
+        super(parent, "Update Available",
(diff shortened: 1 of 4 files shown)


RUGMJ:
Only just realised my auto formatter changed some things. if this is an issue I can revert those changes

src/main/java/tools/redstone/redstonetools/mixin/UpdatePopupMixin.java
```diff
@@ -0,0 +1,68 @@
...
+            return;
+
+        try {
+            String currentVersion = RedstoneToolsClient.MOD_VERSION;
+
+            if (currentVersion.split("-")[2] == "beta")
+                return;
+
+            OkHttpClient client = new OkHttpClient();
+            Request request = new Request.Builder()
```
Matthias1590, line 45: I think we might want to move the request stuff into a different thread so it doesn't freeze the ui, make sure to test without an internet connection as well

src/main/java/tools/redstone/redstonetools/mixin/UpdatePopupMixin.java
```diff
@@ -0,0 +1,68 @@
...
+
+            Gson gson = new Gson();
+            JsonObject release = gson.fromJson(responseBody, JsonObject.class);
+            String newVersion = release.get("tag_name").getAsString();
+
+            if (currentVersion.equals(newVersion))
+                return;
+
+            MinecraftClient.getInstance().setScreen(new UpdatePopupScreen(this));
+        } catch (JsonSyntaxException e) {
```
Matthias1590, line 59: These 2 catch clauses can be ored together `catch (JsonSyntaxException | IOException e)`

src/main/java/tools/redstone/redstonetools/gui/UpdatePopupScreen.java
```diff
@@ -0,0 +1,26 @@
...
+                "An update is available for redstone tools! You are on version {currentVersion} but version {newVersion} is available.");
+    }
+
+    @Override
+    protected void addButtons(int y) {
+        this.addDrawableChild(new ButtonWidget(this.width / 2 + 5, y, 150, 20, Text.of("Go to release"), (button) -> {
+            this.close();
+        }));
+        this.addDrawableChild(new ButtonWidget(this.width / 2 - 155, y, 150, 20, Text.of("Ignore"), (button) -> {
+            RedstoneToolsClient.LOGGER.info("Ignored Update");
```
Matthias1590, line 21: Don't think this needs an info log or a log at all

Matthias1590 (changes requested):

src/main/java/tools/redstone/redstonetools/gui/UpdatePopupScreen.java
RUGMJ: My bad that was leftover from debugging

src/main/java/tools/redstone/redstonetools/mixin/UpdatePopupMixin.java
RUGMJ: That's what the trycatch blocks are for but I haven't actually tested that

RUGMJ: The ui should freeze, it would be strange for the pop up to show before the title screen sometimes and a second after at others

Matthias1590: Fair, I guess it's fine if we have a small timeout on the request, maybe like .25 seconds

RUGMJ: Sounds good

src/main/java/tools/redstone/redstonetools/mixin/UpdatePopupMixin.java
```diff
@@ -24,6 +25,8 @@
 @Mixin(TitleScreen.class)
 
 public class UpdatePopupMixin extends Screen {
+    private static long timeout = 250;
```
Matthias1590, line 28: Could be a constant but this works

Matthias1590 (approved):

PR #159, opened 2023-05-03
Adds a success check to the update request

RUGMJ:


src/main/java/tools/redstone/redstonetools/mixin/UpdatePopupMixin.java
```diff
@@ -54,6 +54,8 @@ public void init(CallbackInfo ci) {
                     .build();
             Response response = client.newCall(request).execute();
             String responseBody = response.body().string();
```
Matthias1590, line 56: The response body should be closed but tbh its only being opened once so probably wont leak more than a megabyte of memory at most

Matthias1590 (approved):

PR #171, opened 2023-05-11
Implemented macro changes

MiranCZ:


src/main/java/tools/redstone/redstonetools/macros/KeyBindingMixin.java
```diff
@@ -0,0 +1,9 @@
+package tools.redstone.redstonetools.macros;
+
+import net.minecraft.client.option.KeyBinding;
+
+public interface KeyBindingMixin {
```
Matthias1590, line 5: Why is there an interface for this?

src/main/java/tools/redstone/redstonetools/macros/Macro.java
```diff
@@ -96,6 +97,11 @@ public Macro createCopy() {
         return new Macro(name,enabled,key,new ArrayList<>(actions),this);
     }
 
+    public void close(){
+        keyBinding.setBoundKey(InputUtil.UNKNOWN_KEY);
+        ((KeyBindingMixin)keyBinding).removeKeybinding(keyBinding);
```
Matthias1590, line 102: Can we use a utils class instead of adding functionality to the keybinding class?

src/main/java/tools/redstone/redstonetools/macros/MacroManager.java
```diff
@@ -173,4 +173,11 @@ private Action getActionFromJson(JsonObject actionJson) {
 
         throw new RuntimeException("Unknown action type: " + type);
     }
+
+    public void updateMacroKeys() {
```
Matthias1590, line 177: What does this macro do? Setting a variable to its current value shouldnt do anything

src/main/java/tools/redstone/redstonetools/macros/gui/commandsuggestor/WorldlessCommandSuggestor.java
```diff
@@ -31,5 +31,13 @@ public static int getY(Object object){
         return yMap.get(object);
     }
 
+    @Override
+    public void refresh() {
+        try{
+            super.refresh();
+        } catch (Exception ignored) {
```
Matthias1590, line 38: In what cases would an exception be thrown? If there's a good reason to catch a certain exception it might be better to catch that specific exception class instead of `Exception`

src/main/java/tools/redstone/redstonetools/macros/gui/widget/commandlist/CommandEntryPlaceHolder.java
```diff
@@ -11,6 +11,7 @@ public CommandEntryPlaceHolder(MinecraftClient client, CommandListWidget owner,
 
     @Override
     public void render(MatrixStack matrices, int index, int y, int x, int entryWidth, int entryHeight, int mouseX, int mouseY, boolean hovered, float tickDelta) {
+        command.setSuggestion("Add new command");
```
Matthias1590, line 14: Does this placeholder go away when the field is focused or not empty like with the macro name field?

src/main/java/tools/redstone/redstonetools/macros/Macro.java
```diff
@@ -96,6 +97,11 @@ public Macro createCopy() {
         return new Macro(name,enabled,key,new ArrayList<>(actions),this);
     }
 
+    public void close(){
+        keyBinding.setBoundKey(InputUtil.UNKNOWN_KEY);
```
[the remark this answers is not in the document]
Matthias1590: Also I think unregisterActions might be a better name than close

src/main/java/tools/redstone/redstonetools/macros/Macro.java
```diff
@@ -27,8 +28,9 @@ public static Macro buildEmpty() {
 
     public Macro(String name, boolean enabled, Key key, List<Action> actions) {
         this(name,enabled,key,actions,null);
-        keyBinding = new InvisibleKeyBinding("macro." + System.nanoTime(),-1,"macros");
+        keyBinding = new KeyBinding("macro." + System.nanoTime(),-1,"macros");
         addKeyBindingToOptions();
+        setKey(key);
```
Matthias1590, line 33: why does this call setKey? doesnt it already call another constructor which sets key?

src/main/java/tools/redstone/redstonetools/macros/gui/commandsuggestor/WorldlessCommandSuggestor.java
Matthias1590: What if somethings wrong and the game cant operate normally anymore? If the crash is caused by rst we should fix it, if it isnt we shouldnt catch it

src/main/java/tools/redstone/redstonetools/macros/KeyBindingMixin.java
Matthias1590: I don't see why you would need an interface for that though, can't you just cast to the implementation?

src/main/java/tools/redstone/redstonetools/macros/Macro.java
Matthias1590: Yeah, maybe just make key private if it isnt and create getters and setters and then call registerKeyBind in the setter so other classes have the option to overwrite the key or to just register the keybind

src/main/java/tools/redstone/redstonetools/macros/Macro.java
Matthias1590: Could you make a utils class for it that uses this line of code internally? something like `KeyBindingUtils.removeKeyBinding(keyBinding)` looks more readable than `((KeyBindingMixin)keyBinding).removeKeybinding(keyBinding)` imo

src/main/java/tools/redstone/redstonetools/macros/KeyBindingMixin.java
Matthias1590: I see, could you add an L to the end of the implementation name then so it becomes Impl instead of Imp?

src/main/java/tools/redstone/redstonetools/macros/MacroManager.java
Matthias1590: This would just be able to call registerKeyBinding on each macro after that method gets added right?

src/main/java/tools/redstone/redstonetools/macros/Macro.java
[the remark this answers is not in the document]
Matthias1590: Oh yeah my bad, this isnt for actions, its for the keybind

src/main/java/tools/redstone/redstonetools/macros/Macro.java
Matthias1590: in this case, imo, yes, `((KeyBindingMixin)keyBinding).removeKeybinding(keyBinding)` isnt very readable

src/main/java/tools/redstone/redstonetools/macros/Macro.java
```diff
@@ -82,8 +80,11 @@ public boolean isCopyOf(Macro macro) {
 
     public void setKey(Key key) {
         this.key = key;
-        if (this.keyBinding != null) {
+        registerKeyBinding();
```
Matthias1590, line 84: Does setting the key using the setKey method unregister the previous key?

src/main/java/tools/redstone/redstonetools/macros/Macro.java
```diff
@@ -41,7 +39,7 @@ public Macro(String name, boolean enabled, Key key, List<Action> actions, Macro
         this.original = original;
     }
 
-    public void addKeyBindingToOptions() {
+    public void addKeyBinding() {
```
Matthias1590, line 43: This seems more like registerKeyBinding than the current registerKeyBinding, might be nice to rename this one to registerKeyBinding and registerKeyBinding to setKeyBindingKeyCode

Matthias1590 (changes requested):

Matthias1590 (approved):

PR #177, opened 2023-05-12
Replaced string serialized block states with NBT serialized states

orbyfied:


src/main/java/tools/redstone/redstonetools/utils/BlockStateNbtUtil.java
```diff
@@ -1,64 +1,174 @@
...
-        for (Property<?> property : blockState.getEntries().keySet()) {
-            valueMap.put(property.getName(), blockState.getEntries().get(property));
+    /**
+     * Serializes the given block state into an NBT compound.
+     *
+     * @param state The state.
+     * @return The NBT tag or null if the block state is null.
+     */
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    public static NbtCompound toNBT(BlockState state) {
```
Matthias1590, line 32: I highly doubt that there isnt code that does this already, might be nice to just use that instead for the blockstate to and from nbt conversion

src/main/java/tools/redstone/redstonetools/features/commands/CopyStateFeature.java
```diff
@@ -38,13 +38,7 @@ protected Either<ItemStack, Feedback> getItemStack(ServerCommandSource source, B
...
     private int addBlockStateNbt(ItemStack itemStack, BlockState blockState) {
         addBlockStateText(itemStack);
-
-        NbtCompound nbt = itemStack.getOrCreateNbt();
-        String stringState = BlockStateNbtUtil.blockStateToString(blockState);
-        if (stringState == null) return -1;
-
-        nbt.putString("blockstate",stringState);
-
+        BlockStateNbtUtil.createPlacementStack(itemStack, blockState);
```
Matthias1590, line 41: Might be nice to either assign the return value here or rename the method to make it clear that it modifies the itemstack you pass

src/main/java/tools/redstone/redstonetools/utils/BlockStateNbtUtil.java
```diff
@@ -1,64 +1,174 @@
...
+     * Modifies or creates an item stack which should place the exact given
+     * block state when attempted.
+     *
+     * If the item stack has to be created it has a count of 1 by default.
+     *
+     * @param stack The stack to modify. Null signals a new stack.
+     * @param state The placement state.
+     * @return The item stack.
+     */
+    public static ItemStack createPlacementStack(ItemStack stack, BlockState state) {
```
Matthias1590, line 123: Should either create or modify an itemstack, not both

Matthias1590 (changes requested):

src/main/java/tools/redstone/redstonetools/utils/BlockStateNbtUtil.java
orbyfied: I looked into the chunk serialization and couldn't find any reusable code.
This works though, so I suggest we just keep it in.

src/main/java/tools/redstone/redstonetools/utils/BlockStateNbtUtil.java
Matthias1590, line 123: Should probably either create an object or modify one that's passed in, this violates srp

Matthias1590 (changes requested):

Matthias1590 (approved):

PR #214, opened 2023-06-04
Replace javax.json-api (and its implementation) with builtin Gson

xtrm-en:
### 228kb baby! :champagne: 

This commit removes the usage of a third-party dependency for Json parsing/writing and replaces its usages with Minecraft's provided [Gson](https://github.com/google/gson/) library.

This does not only refactors usages but also improves them by using Gson's binding functionality (instantiating a class automatically based on Json string values) via custom adapter classes (located in `tools.redstone.redstonetools.macros.gson`) that make it subjectively easier to write and read Json "translation" code. 

This also adds the nice convenience of pretty printing to file, meaning indented, humanly readable Json formatting.


JukoiDev (dismissed):
LGTM

src/main/java/tools/redstone/redstonetools/macros/MacroManager.java
```diff
@@ -23,29 +21,32 @@ public class MacroManager {
...
+                        }.getType()
+                );
+                if (parsedMacros != null) {
+                    Collections.addAll(this.macros, parsedMacros);
+                }
             }
-        } catch (Exception e) {
-            e.printStackTrace();
+        } catch (Throwable t) {
+            throw new RuntimeException("Couldn't load macros config", t);
```
Matthias1590, line 44: not being able to parse the macros shouldnt crash the game

src/main/java/tools/redstone/redstonetools/macros/MacroManager.java
```diff
@@ -23,29 +21,32 @@ public class MacroManager {
...
         }
 
-        if (macrosJson == null) {
-            macros = new ArrayList<>();
-            macros.addAll(getDefaultMacros());
-        } else {
-            macros = getMacrosFromJson(macrosJson);
+        if (this.macros.isEmpty()) {
```
Matthias1590, line 47: this would cause the default macros to be added the next time the game launches after the user deletes them, i think it makes more sense to only load default macros if the macro file could not be read

src/main/java/tools/redstone/redstonetools/macros/MacroManager.java
Matthias1590: just noticed thats what it does in the older version as well, my bad, it probably shouldnt crash though

src/main/java/tools/redstone/redstonetools/macros/gson/ActionAdapter.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class ActionAdapter implements JsonSerializer<Action>, JsonDeserializer<Action> {
+    @Override
+    public JsonElement serialize(Action action, Type type, JsonSerializationContext context) {
+        JsonObject object = new JsonObject();
+        if (action instanceof CommandAction commandAction) {
+            object.addProperty("type", "command");
+            object.addProperty("command", commandAction.command);
+        } else {
+            throw new RuntimeException("Unknown action type: " + action.getClass().getName());
```
Matthias1590, line 17: same here, i dont think we should crash

src/main/java/tools/redstone/redstonetools/macros/gson/ActionAdapter.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    @Override
+    public Action deserialize(JsonElement jsonElement, Type type, JsonDeserializationContext context) throws JsonParseException {
+        JsonObject object = jsonElement.getAsJsonObject();
+        String typeStr = object.get("type").getAsString();
+        if (typeStr.equals("command")) {
+            return new CommandAction(object.get("command").getAsString());
+        }
+        throw new RuntimeException("Unknown action type: " + typeStr);
```
Matthias1590, line 29: same here, dont think we should crash

src/main/java/tools/redstone/redstonetools/telemetry/TelemetryManager.java
```diff
@@ -1,67 +1,62 @@
...
-        } catch (Exception e) {
-            e.printStackTrace();
+            telemetryFilePath = MinecraftClient.getInstance().runDirectory.toPath()
+                    .resolve("config")
+                    .resolve("redstonetools")
+                    .resolve("telemetry.json");
+
+            loadSettingsFromJson();
+        } catch (Throwable t) {
+            throw new RuntimeException("Coudln't load telemetry config", t);
```
Matthias1590, line 29: same crashing thing here

Matthias1590 (changes requested):

src/main/java/tools/redstone/redstonetools/macros/MacroManager.java
```diff
@@ -17,20 +19,27 @@
...
         this.macros = new ArrayList<>();
-
         try {
             macrosFilePath = MinecraftClient.getInstance().runDirectory.toPath()
                     .resolve("config")
                     .resolve("redstonetools")
                     .resolve("macros.json");
+        } catch (Throwable t) {
+            LOGGER.fatal("Couldn't load macros config path", t);
+            throw new RuntimeException("Couldn't load macros config path", t);
```
Matthias1590, line 35: I dont think we should crash

src/main/java/tools/redstone/redstonetools/telemetry/TelemetryManager.java
```diff
@@ -23,10 +23,15 @@ public TelemetryManager() {
                     .resolve("config")
                     .resolve("redstonetools")
                     .resolve("telemetry.json");
+        } catch (Throwable t) {
+            LOGGER.fatal("Couldn't load telemetry config path", t);
+            throw new RuntimeException("Couldn't load telemetry config path", t);
```
Matthias1590, line 28: Don't think we should crash here either

Matthias1590 (changes requested):

src/main/java/tools/redstone/redstonetools/macros/MacroManager.java
xtrm-en: Those crashes are redundant ones, basically where Minecraft doesn't get initialized properly, which shouldn't happen (on a release build). Additionally, if this file isn't properly found, the rest of the code would be the one to crash due to NPEs, so early throwing would be the best course of actions here.

src/main/java/tools/redstone/redstonetools/telemetry/TelemetryManager.java
xtrm-en: Same as previous comment

Matthias1590 (approved):

Matthias1590 (approved):

Matthias1590 (approved):

Matthias1590:
Yeah so were gonna refactor everything and idk if it would be worth the testing to merge this, sorry

Issue #216, opened 2023-06-04
Unexpected Behavior with Macro Binding


Bitlett:
If two macros are bound to the same key, the mod will ALWAYS pick the one that was created last. Even if that one is disabled.

[Discord Thread](1114887642476183603)

JayMensink:
pls assign me 🥺

JayMensink:
BUG FOUND: Re-opening a macro will force the text to be red. This happens because the detector for duplicate bindings detects the macro and concludes that the bind is in use.

PR #218, opened 2023-06-04
Duplicate Macro Bind Indicator

JayMensink:
Issue #216

Solution: If you try to bind a macro to a key that is already bound, it will indicate this by making the bind key red.


src/main/java/tools/redstone/redstonetools/macros/Macro.java
```diff
@@ -85,10 +86,12 @@ public void setKey(Key key) {
     }
 
     public void changeKeyBindingKeyCode() {
+
```
xtrm-en, line 89: why the spaces :sob:

xtrm-en (approved):
lgtm

src/main/java/tools/redstone/redstonetools/macros/Macro.java
JayMensink: i was messing around and apparently i left these in on accident after adding and removing some code

src/main/java/tools/redstone/redstonetools/utils/KeyBindingUtils.java
```diff
@@ -9,4 +11,12 @@ public static void removeKeyBinding(KeyBinding keyBinding) {
         ((KeyBindingMixin)keyBinding).removeKeybinding(keyBinding);
     }
 
+    public static boolean isKeyAlreadyBound(Key key) {
+        KeyBinding[] keyBindings = MinecraftClient.getInstance().options.allKeys;
+        for (KeyBinding keyBinding : keyBindings) {
+            if (keyBinding.getDefaultKey() == key) { return true; }
```
Matthias1590, line 17: Isn't the default key hardcoded and not the one set by the user?

JayMensink: ugh you're right time to fix that

JayMensink: Successfully changed.

Matthias1590 (approved):

Matthias1590 (approved):

JayMensink:
@Matthias1590 can you merge this

PR #219, opened 2023-06-04
Improve actions workflow.

xtrm-en:
This currently adds a use to [`actions/cache@v3`](https://github.com/actions/cache) for faster build times, as well as
[`actions/upload-artifact@v3`](https://github.com/actions/upload-artifact) for easier testing of incoming PRs.

This also launches Gradle with `--no-daemon` which speeds it up just a bit which is pretty cool.

I'm noticing that I unknowingly removed
```yml
permissions:
  contents: read
```
and I'm unsure of what it's supposed to do, the [official documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) doesn't really do it justice.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -4,10 +4,6 @@ name: Build & test
 
 on:
   pull_request:
-  workflow_dispatch:
-  push:
-    branches-ignore:
-      - main
 
 concurrency:
   group: ${{ github.head_ref || format('{0}-{1}', github.ref, github.run_number) }}
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -21,17 +21,17 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           fetch-depth: 10
 
       - name: Set up JDK 17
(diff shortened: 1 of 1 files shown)


Matthias1590:
Pretty sure setting the permission to read would override the github token perms to make it only able to read from the github api and not write by sending post requests and such

Matthias1590:
What does this PR change exactly?

xtrm-en:
> What does this PR change exactly?

Pretty much what's said the OP, faster build times, allows for artifact publishing

It should also allow for PR actions to be run directly when the PR author is a contributor I believe? Not sure if that's a workflow thing or a repo settings thing.

xtrm-en:
Example of artifact builds at the end of the page: https://github.com/RedstoneTools/redstonetools-mod/actions/runs/5169729816?pr=219

Matthias1590:
Ah ok looks good then, is this still a draft?

xtrm-en:
I was thinking of keeping it a draft until more input is given but I don't know if that's necessary

Matthias1590:
Do we need the push to main? You can only add to main using pull requests anyway

Matthias1590 (changes requested):
We can get rid of the workflow_dispatch and push triggers

Matthias1590 (approved):

PR #223, opened 2023-06-04
Introduce Starscript, a text formatting engine.

xtrm-en:
### This is very much a draft

Preview: https://youtu.be/NUOosjiHhFY

This PR adds support for "scripting"-like formatting in commands, which would allow for interesting scenarios and more uses of macros.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,6 +27,10 @@ repositories {
 		name = "JitPack"
 		url = "https://jitpack.io"
 	}
+	maven {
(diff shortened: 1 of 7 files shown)


Matthias1590:
I dont think we'd be able to use starscript if we end up using marshal, I think it'd be better if we built something into marshal ourselves, that way we can still use autocompletion and such, I do like the idea of being able to evaluate expressions in macros though

PR #224, opened 2023-06-04
Languages update

ikyynites:
Added translation keys for new languages, consulted friends who are fluent in these languages to check that translations were accurate and correct.


src/main/resources/assets/redstonetools/lang/fr_fr.json
```diff
@@ -0,0 +1,6 @@
+{
+    "gamerule.doContainerDrops": "Effectuer des largages de conteneurs",
```
xtrm-en, line 2: This is really weirdly formulated, and doesn't follow Minecraft's official translation.

I'd suggest "Butin des conteneurs", and as a description: "Contrôle les butins des conteneurs"

xtrm-en (changes requested):

src/main/resources/assets/redstonetools/lang/fr_fr.json
ikyynites: it is, my friends said it was hard to translate without context, and i cant verify their translations myself. it was taking me ages to get the official translations because i couldnt find any reliable sources other than looking up the hashed names of each individual language and finding it in the minecraft assets. ill give them all another look tomorrow

ikyynites: ive just removed translations that i cant verify. will re-add them once i can cross-reference with the official translations for other gamerules

ikyynites: as of now, i would push this and add the couple languages that have been properly translated. and ill add a few more later on.

Issue #226, opened 2023-06-04
//read default blockstate doesnt work


Matthias1590:
You cant run //read without explicitly setting the blockstate

https://discord.com/channels/1071119659346169907/1115012922528251964

Matthias1590:
Fixed by #228

PR #228, opened 2023-06-05
Fixed //read default blockstate

Bani-OK:
Fixed bug #226

`BlockStateArgument` requires a set of properties to be passed upon creation. `LIT_LAMP_ARG` - default blockstate for //read command was created without this set and used afterwards.
I have added properties to the initialization of the default block state.


JayMensink:
Please comment on the original issue asking to be assigned next time so that people can see that you're working on it.

Issue #229, opened 2023-06-05
Macro Command - Incorrect Max Length


JayMensink:
If the length of a macro command exceeds that of the chat box's character limit, the command will simply cut off. The macro command input box should have a character limit to keep this from happening.

[Discord Thread](https://discord.com/channels/1071119659346169907/1114917852235976745)

JayMensink:
I'll get this done. Assign me pls.

JayMensink:
Fix implemented.
Mark [The Forum Post](https://discord.com/channels/1071119659346169907/1114917852235976745) as Implemented pls @Matthias1590

PR #234, opened 2023-06-05
Macro command length

JayMensink:
If the length of a macro command exceeds that of the chat box's character limit, the command will simply cut off. The macro command input box has had it character limit reduced to 255 to keep this from happening.

Issue #229


JayMensink:
Are you kidding me dude I even told it to clone from the dev branch

PR #235, opened 2023-06-05
Macro Command Max Length Fix

JayMensink:
If the length of a macro command exceeds that of the chat box's character limit, the command will simply cut off. The macro command input box has had it character limit reduced to 255 to keep this from happening.

Issue #229


JayMensink:
> Wouldnt it be better to execute commands in a different way? (I didnt look into it, but there seems to be something like `CommandDispatcher.execute()`)

Not really. When a `CommandAction` is run, it sends the server a chat packet with the command. This is precisely how it would go if the command were typed in via chat. I already looked into it, and there's no packet type dedicated to commands, which means there's also no way to get macros to run a command longer than 256 characters.

Matthias1590:
99% sure this would still break if the user doesnt enter the first / since if they dont we add one which could result in a 256 long command but fuck that im merging this

PR #236, opened 2023-06-05
Added dutch 'translations'

Miniontoby:
Added dutch 'translations' where I say 'translations' since it isn't really translating the whole thing (yet)


orbyfied:
I think we should wait to merge any translations until we have the framework to translate more in place, that should be done pretty soon though.

Miniontoby:
> 

Maybe this can help: https://github.com/gbl/CrowdinTranslate

PR #239, opened 2023-06-05
Fixed copystate crashing the game when used on TileEntities

Javatrix912:
Sorry it took so long to pull, I'm new to GitHub.


Matthias1590:
Not sure if this solves it but it turns out the bug was caused by an accessor, renaming it fixed it

Javatrix912:
Well, I had fun at least :)
So you say that the NAME of MinecraftClientAccessor cause an error? Or the method? I would have never thought that can be an issue. Thanks for letting me contribute though!

PR #240, opened 2023-06-05
/itembind and compatibility fix

### LINKED ISSUE #217: Remove text highlighting from the main menu in the macro gui
Highlighting in the macro gui is cool, but it causes a lot of problems, so its best to just remove it from the main menu and have it only, when player is in a world

MiranCZ:
resolves #217


src/main/java/tools/redstone/redstonetools/utils/ItemUtils.java
```diff
@@ -1,23 +1,28 @@
...
+        NbtCompound displayNbt = itemStack.getSubNbt("display");
+        NbtList loreList = new NbtList();
 
-    static {
-        for (Item item : Registry.ITEM) {
-            ITEM_MAP.put(item.toString(), item);
+        if (displayNbt == null) {
+            displayNbt = new NbtCompound();
+        } else {
+            loreList = (NbtList) displayNbt.get("Lore");
```
Matthias1590, line 20: Maybe add a null check and return if displayNbt.get("Lore") is null just to be sure, casting would have to happen after the null check

src/main/java/tools/redstone/redstonetools/utils/ItemUtils.java
```diff
@@ -1,23 +1,28 @@
...
         }
-    }
 
-    public static Item getItemByName(String itemName) {
-        return ITEM_MAP.getOrDefault(itemName,null);
+        loreList.add(NbtString.of("\"(+"+text +")\""));
```
Matthias1590, line 23: What would happen if you change the bind of an item? Would it add multiple (+Command)? Cause it should replace them

src/main/java/tools/redstone/redstonetools/mixin/features/ItemBindMixin.java
```diff
@@ -0,0 +1,71 @@
...
+
+        @Inject(method = "useOnBlock", at = @At("HEAD"), cancellable = true)
+        public void checkCommandNBT(ItemUsageContext context, CallbackInfoReturnable<ActionResult> cir) {
+            if (tryToExecuteNBTCommand(context.getHand(), context.getWorld())) {
+                cir.setReturnValue(ActionResult.PASS);
+            }
+        }
+
+        private boolean tryToExecuteNBTCommand(Hand hand, World world) {
+            if (hand == Hand.OFF_HAND || world.isClient) return false;
```
Matthias1590, line 47: What does the world.isClient check do?

src/main/java/tools/redstone/redstonetools/features/commands/ItemBindFeature.java
```diff
@@ -0,0 +1,63 @@
...
+        if (!player.getInventory().contains(bindStack)) {
+            bindStack = null;
+            waitingForCommand = false;
+            return false;
+        }
+
+        bindStack.getOrCreateNbt().put("command", NbtString.of(command));
+        ItemUtils.addExtraNBTText(bindStack,"Command");
+
+        player.sendMessage(new LiteralText("[RST] Successfully added: '" + command + "' to this item (" + bindStack.getItem().toString() + ")!"),false);
```
Matthias1590, line 55: "Successfully bound" makes more sense imo since we're not adding multiple ones

src/main/java/tools/redstone/redstonetools/features/commands/ItemBindFeature.java
```diff
@@ -0,0 +1,63 @@
...
+    @Override
+    protected Feedback execute(ServerCommandSource source) throws CommandSyntaxException {
+
+        ServerPlayerEntity player = source.getPlayer();
+        ItemStack mainHandStack = player.getMainHandStack();
+        if (mainHandStack == null || mainHandStack.getItem() == Items.AIR) {
+            return Feedback.error("You need to be holding an item!");
+        }
+        if (mainHandStack.getNbt() != null && mainHandStack.getNbt().contains("command")) {
+            return Feedback.error("This item already has a command bind to it!");
```
Matthias1590, line 35: I think it makes more sense to override the bound command instead of failing here

Matthias1590 (changes requested):

src/main/java/tools/redstone/redstonetools/mixin/features/ItemBindMixin.java
Matthias1590: Ah ok, cool

src/main/java/tools/redstone/redstonetools/utils/ItemUtils.java
Matthias1590: Yeah i think thats my fault, i closed the itembind pr thinking this was just that one but with the macro stuff added

src/main/java/tools/redstone/redstonetools/features/commands/ItemBindFeature.java
```diff
@@ -0,0 +1,61 @@
...
+        if (!player.getInventory().contains(bindStack)) {
+            bindStack = null;
+            waitingForCommand = false;
+            return false;
+        }
+
+        bindStack.getOrCreateNbt().put("command", NbtString.of(command));
+        ItemUtils.addExtraNBTText(bindStack,"Command");
+
+        player.sendMessage(new LiteralText("[RST] Successfully added: '" + command + "' to this item (" + bindStack.getItem().toString() + ")!"),false);
```
Matthias1590, line 53: "added:" should be "bound command"

src/main/java/tools/redstone/redstonetools/features/commands/ItemBindFeature.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        bindStack = mainHandStack;
+        waitingForCommand = true;
+
+
+        return Feedback.success("Please run the command you want to bind to this item (" + mainHandStack.getItem().toString()+")");
+    }
+
+    public static boolean addCommand(ClientPlayerEntity player, String command) {
```
Matthias1590, line 42: This should return feedback

Matthias1590 (changes requested):

src/main/java/tools/redstone/redstonetools/features/commands/ItemBindFeature.java
Matthias1590: You could make it return a nullable feedback and then use the feedback sender in the mixin to send the feedback

Matthias1590 (approved):

Matthias1590:
Leaving a world does not cause it to stop listening for a command

Matthias1590:
> > Leaving a world does not cause it to stop listening for a command
> 
> And what happens if you join and execute the command?

It will send the success message but doesnt actually bind the command

Matthias1590 (changes requested):
Still has some issues with /itembind

Matthias1590:
Might just be me but /itembind just doesnt seem to listen for a command anymore.

Steps to reproduce:
- Enter a world
- Get out an item
- Run `/itembind`
- Run `/say hello`

You'll see "hello" in chat because the say command just executed instead of getting bound, when you try to right click the item it wont do anything and it wont give you any feedback either.

Matthias1590:
might just be me, but it completely stopped working, also could you maybe use the new formatting for the feedback?

Issue #238, opened 2023-06-05
Replacing the /glass command with /colored


maghedo243:
This would have the same functionality as /glass, but it would be argument-based and allow the user to choose what they are material they are receiving. This would also add concrete and possibly terracotta to the mix instead of just glass and wool.

Matthias1590:
This should replace glass

maghedo243:
Understood

PR #244, opened 2023-06-05
/glass -> /colored 

maghedo243:
These changes turn the GlassFeature.java into the new and improved ColoredFeature.java. The extra files are all involved with the /colored feature. Issue #238 

The material options:
- Glass
- Wool
- Concrete
- Terracotta
- Glazed Terracotta


src/main/java/tools/redstone/redstonetools/features/commands/ColoredFeature.java
```diff
@@ -0,0 +1,86 @@
...
+
+@AutoService(AbstractFeature.class)
+@Feature(name = "Colored", description = "Gives the player specified variant of block being looked at, with the same color. Default is White.", command = "colored")
+public class ColoredFeature extends PickBlockFeature {
+    @Override
+    protected boolean requiresBlock() {
+        return false;
+    }
+
+    public static final Argument<ColoredBlockArgument> blockType = Argument.ofType(material());
```
Matthias1590, line 30: Could you move the field declarations to the top of the file?

src/main/java/tools/redstone/redstonetools/utils/ColoredBlockArgument.java
```diff
@@ -0,0 +1,14 @@
+package tools.redstone.redstonetools.utils;
+
+public enum ColoredBlockArgument {
```
Matthias1590, line 3: Could probably be renamed to ColoredMaterial, it's not really related to arguments and it doesn't really represent a block either, ig ColoredBlockType could work too in which case the serializer should probably be called ColoredBlockTypeSerializer and the method that returns instance would be something like coloredBlockType()

src/main/java/tools/redstone/redstonetools/features/commands/ColoredFeature.java
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        var coloredBlock = getColorVariant(blockInfo.block);
+
+        return Either.left(coloredBlock == null
+            ? new ItemStack(Blocks.GLASS)
+            : new ItemStack(coloredBlock.toBlock()));
+    }
+
+    private ColoredBlock getColorVariant(Block block) {
```
Matthias1590, line 51: How is this different from ColoredBlock.fromBlock()?

Matthias1590 (changes requested):

src/main/java/tools/redstone/redstonetools/features/commands/ColoredFeature.java
maghedo243: Well I was just editing the GlassFeature code. ".fromBlock()" appears to get a blockId that is independent of the actual color that the block is. And my "getColorVariant" actually chooses the color based on the argument.

src/main/java/tools/redstone/redstonetools/utils/ColoredBlockArgument.java
maghedo243: The only reason I named it ColoredBlockArgument was to copy the naming convention of the DirectionArgument, but I'll change it to your suggestions.

src/main/java/tools/redstone/redstonetools/features/commands/ColoredFeature.java
maghedo243: Yep

src/main/java/tools/redstone/redstonetools/utils/ColoredBlockTypeArgument.java
```diff
@@ -1,6 +1,6 @@
 package tools.redstone.redstonetools.utils;
 
-public enum ColoredBlockArgument {
+public enum ColoredBlockTypeArgument {
```
Matthias1590, line 3: I dont think this should have Argument in its name since its just a colored block type, nothing more than that

Matthias1590 (changes requested):

src/main/java/tools/redstone/redstonetools/utils/ColoredBlockTypeArgument.java
maghedo243: I guess it's technically not specifically for the arguments, so I'll change it

Matthias1590:
You didnt change anything? Are you sure you pushed your changes?

maghedo243:
> You didnt change anything? Are you sure you pushed your changes?

Pretty sure yeah. I can see them here

Matthias1590:
Wait what I think github wasnt showing them, my bad

Matthias1590:
The same happens when looking at a non colored block such as oak stairs

Matthias1590:
Doesnt seem to work when looking at any colored block either

Matthias1590:
Yeah my bad this works perfectly fine

PR #262, opened 2023-06-06
Feedback now allows for values and these will be formatted

InvertedOwl:
Feedback now has an optional values param. To format the value just put %s. As well as the rst tag being all good-looking now

### DIFF
--- a/src/main/java/tools/redstone/redstonetools/features/feedback/Feedback.java
+++ b/src/main/java/tools/redstone/redstonetools/features/feedback/Feedback.java
@@ -6,19 +6,32 @@ import javax.annotation.Nullable;
 
 public abstract class Feedback {
     private final @Nullable String message;
+    private final @Nullable String[] values;
 
     protected Feedback(@Nullable String message) {
         this.message = message;
+        this.values = new String[0];
+    }
+
+    protected Feedback(@Nullable String message, @Nullable String[] values) {
+        this.message = message;
+        this.values = values;
     }
 
     private String formatMessage(String message) {
-        return "[RST] " + message;
+        return String.format("%s[%sRST%s]%s ", Formatting.GRAY, Formatting.RED, Formatting.GRAY, Formatting.RESET) + message;
     }
 
     public final String getMessage() {
-        return formatMessage(message != null
-                ? message
-                : getDefaultMessage());
+        if (message != null) {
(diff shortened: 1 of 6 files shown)


src/main/java/tools/redstone/redstonetools/features/feedback/Feedback.java
```diff
@@ -6,19 +6,32 @@
...
 public abstract class Feedback {
     private final @Nullable String message;
+    private final @Nullable String[] values;
 
     protected Feedback(@Nullable String message) {
         this.message = message;
+        this.values = new String[0];
+    }
+
+    protected Feedback(@Nullable String message, @Nullable String[] values) {
```
Matthias1590, line 16: You could use varargs instead so you dont have to create an array everytime

src/main/java/tools/redstone/redstonetools/features/feedback/Feedback.java
```diff
@@ -53,11 +66,17 @@ public FeedbackType getType() {
     public static Success success(@Nullable String message) {
```
Matthias1590, line 66: This could also use varargs which would remove the need for the next method

src/main/java/tools/redstone/redstonetools/features/feedback/Feedback.java
```diff
@@ -53,11 +66,17 @@ public FeedbackType getType() {
...
         return new Success(message);
     }
+    public static Success success(@Nullable String message, @Nullable String[] values) {
+        return new Success(message, values);
+    }
 
     private static class Success extends Feedback {
         public Success(@Nullable String message) {
```
Matthias1590, line 74: Same here, if you use varargs you wont need the next method

Matthias1590 (changes requested):

src/main/java/tools/redstone/redstonetools/features/feedback/Feedback.java
maghedo243: You can do that?

src/main/java/tools/redstone/redstonetools/features/feedback/Feedback.java
InvertedOwl: Great point

Matthias1590 (approved):

Issue #268, opened 2023-06-07
Implementing //write


DreiMu:
Implementing https://discord.com/channels/1071119659346169907/1106312444353204244 so you can use //write like //read but writing to an input instead of reading an output

PR #269, opened 2023-06-07
Implemented simple //write for binary values (#268)

DreiMu:
Simple implementation for a write command, still needs testing, only bug found for now is the redstonelamps turning off, if there is one updated next to them


Matthias1590:
The default on block should be a redstone block, the fact that lamps turn off when updated is normal

Matthias1590:
Looks good, lmk when you're done editing & testing

DreiMu:
I can't find any more errors and have nothing more to implement, somebody else should still test it though

Matthias1590 (approved):

Matthias1590:
image
Not entirely sure whats going on here, it seems to be writing more bits than would fit, if I make the offset 1 it works, maybe it should have a bitsize instead of offset and then just calculate the offset manually

Matthias1590 (changes requested):
Still has some issues, see comment above

DreiMu:
They should be fixed now

src/main/java/tools/redstone/redstonetools/features/commands/BinaryBlockWriteFeature.java
```diff
@@ -0,0 +1,100 @@
...
+            null
+    );
+
+    private static final BlockStateArgument DEFAULT_OFF_ARG = new BlockStateArgument(
+            Blocks.AIR.getDefaultState(),
+            Collections.emptySet(),
+            null
+    );
+
+    public static final Argument<Integer> offset = Argument
```
Matthias1590, line 38: instead of an offset, take the bitsize of the number, the number should probably come before this and then the bitsize

Matthias1590 (changes requested):

src/main/java/tools/redstone/redstonetools/features/commands/BinaryBlockWriteFeature.java
DreiMu: Is implemented, if it worked before, it should still work

JayMensink:
Please mention the issue next time so your PR shows up on the issue.

#268

DreiMu:
I had the issue in the name, so that doesn't work, so I'll do it next time in the comment

DreiMu:
fixed both problems

Matthias1590:
This behaves rather weirdly imo, when you select 7 blocks (b a b a b a b, where b is a block and a is air) and try to write a 4 bit number it'll tell you the selection isnt a multiple of the bit count when I think its a perfectly valid input. I think we're gonna have to postpone this feature and add it after the refactor

DreiMu:
It was pretty easy to fix, if you want to test it again, it should now work

Matthias1590:
Not sure whether this is an issue with this feature or not but could you look into this?
image

DreiMu:
This seems to be a problem of the feedback formatting, the only thing I could do about it, is to shorten the text

siebekorf-cell (approved):
Nice feature

PR #270, opened 2023-06-08
updated gradle dependencies

JohanVonElectrum:
This pull request is intended to facilitate the process for anyone who wants to help with the port, serving as a basis to start with the dependencies already updated for 1.19.4.


Matthias1590:
The 1.19 port branch is kinda just a placeholder, I think it would be better to create a new project from scratch if we're porting to 1.19

Issue #246, opened 2023-06-06
Support for /ss with values > 15


JayMensink:
Stacking unstackable items can make a comparator read more than 15 ss from a container.

[Discord Thread](https://discord.com/channels/1071119659346169907/1114600667282145470)

JayMensink:
I'll look into this

JayMensink:
![image](https://github.com/RedstoneTools/redstonetools-mod/assets/108234409/82143e9a-96b2-44ce-a686-e9ba9bd25824)
![image](https://github.com/RedstoneTools/redstonetools-mod/assets/108234409/d482d344-7d30-4da9-8d0d-5a98e473e24b)
![image](https://github.com/RedstoneTools/redstonetools-mod/assets/108234409/b00a7349-1241-4f4c-8179-b77ef420b595)

i got it working lol

PR #271, opened 2023-06-08
ss-command-feature

JohanVonElectrum:
Generalized from barrels to any common signal provider block and for all possible ss.


src/main/java/tools/redstone/redstonetools/features/commands/SsCommandFeature.java
```diff
@@ -0,0 +1,53 @@
...
+import tools.redstone.redstonetools.features.feedback.Feedback;
+import tools.redstone.redstonetools.utils.SignalBlock;
+
+import java.util.Random;
+
+import static tools.redstone.redstonetools.features.arguments.serializers.IntegerSerializer.integer;
+
+@AutoService(AbstractFeature.class)
+@Feature(name = "Signal Strength Block", description = "Creates a block with the specified signal strength.", command = "ss")
+public class SsCommandFeature extends CommandFeature {
```
Matthias1590, line 21: could probably be named SignalStrengthBlockFeature

src/main/java/tools/redstone/redstonetools/features/commands/SsCommandFeature.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private final static SimpleCommandExceptionType INVALID_SIGNAL_STRENGTH = new SimpleCommandExceptionType(Text.of("Invalid signal strength!"));
```
Matthias1590, line 23: This isnt used anywhere

src/main/java/tools/redstone/redstonetools/features/commands/SsCommandFeature.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public static final Argument<Integer> signalStrength = Argument
+            .ofType(integer(0));
+
+    public static final Argument<SignalBlock> block = Argument
```
Matthias1590, line 28: I dont think theres a reason to let the user pick the type of block

src/main/java/tools/redstone/redstonetools/features/commands/SsCommandFeature.java
```diff
@@ -0,0 +1,53 @@
...
+            .ofType(SignalBlockSerializer.signalBlock())
+            .withDefault(SignalBlock.AUTO);
+
+    @Override
+    protected Feedback execute(ServerCommandSource source) throws CommandSyntaxException {
```
Matthias1590, line 33: This doesnt throw a CommandSyntaxException

src/main/java/tools/redstone/redstonetools/features/commands/SsCommandFeature.java
```diff
@@ -0,0 +1,53 @@
...
+        try {
+            source.getPlayer().giveItemStack(block.getValue().getItemStack(signalStrength.getValue()));
```
Matthias1590, line 35: Could this be abstracted? Its rather unreadable like this

src/main/java/tools/redstone/redstonetools/features/commands/SsCommandFeature.java
Matthias1590: Just an if statement should be enough to decide whether to use a barrel or a command block

src/main/java/tools/redstone/redstonetools/utils/SignalBlock.java
```diff
@@ -0,0 +1,39 @@
+package tools.redstone.redstonetools.utils;
+
+import net.minecraft.block.Block;
+import net.minecraft.block.Blocks;
+import net.minecraft.item.ItemStack;
+
+public enum SignalBlock {
```
Matthias1590, line 7: A bit overkill imo, especially since the type of block doesnt really matter

src/main/java/tools/redstone/redstonetools/utils/SignalBlockSupplier.java
```diff
@@ -0,0 +1,101 @@
...
+import net.minecraft.item.Items;
+import net.minecraft.nbt.NbtCompound;
+import net.minecraft.nbt.NbtList;
+import net.minecraft.text.LiteralText;
+import net.minecraft.text.MutableText;
+import net.minecraft.util.Formatting;
+import net.minecraft.util.registry.Registry;
+
+@FunctionalInterface
+public interface SignalBlockSupplier {
```
Matthias1590, line 16: This also wouldnt be necessary with a simple if statement

Matthias1590:
Theres still some unresolved convos

src/main/java/tools/redstone/redstonetools/features/commands/SsCommandFeature.java
```diff
@@ -0,0 +1,53 @@
...
+import tools.redstone.redstonetools.features.arguments.serializers.SignalBlockSerializer;
+import tools.redstone.redstonetools.features.feedback.Feedback;
+import tools.redstone.redstonetools.utils.SignalBlock;
+
+import java.util.Random;
+
+import static tools.redstone.redstonetools.features.arguments.serializers.IntegerSerializer.integer;
+
+@AutoService(AbstractFeature.class)
+@Feature(name = "Signal Strength Block", description = "Creates a block with the specified signal strength.", command = "ss")
```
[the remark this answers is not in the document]
JohanVonElectrum: Sure? It's replacement of previous /ss

src/main/java/tools/redstone/redstonetools/features/commands/SsCommandFeature.java
Matthias1590: My bad

src/main/java/tools/redstone/redstonetools/features/commands/SsCommandFeature.java
Matthias1590: Maybe put some parts into methods that have meaningful names cause rn its kinda hard to see what it actually does

src/main/java/tools/redstone/redstonetools/utils/SignalBlock.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public static SignalBlock getBestBlock(int signal) {
+        if (signal < 1780) return BARREL;
+        else return COMMAND_BLOCK;
+    }
+
+    private final Block block;
+    private final SignalBlockSupplier supplier;
+
+    SignalBlock(Block block, SignalBlockSupplier supplier) {
```
Matthias1590, line 30: Do we need suppliers? Can't we just have some max capacity or something?

src/main/java/tools/redstone/redstonetools/utils/SignalBlock.java
```diff
@@ -0,0 +1,39 @@
...
+    DROPPER(Blocks.DROPPER, SignalBlockSupplier.container_9()),
+    HOPPER(Blocks.HOPPER, SignalBlockSupplier.container_5()),
+    BREWING_STAND(Blocks.BREWING_STAND, SignalBlockSupplier.container_5()),
+    FURNACE(Blocks.FURNACE, SignalBlockSupplier.container_3()),
+    SMOKER(Blocks.SMOKER, SignalBlockSupplier.container_3()),
+    BLAST_FURNACE(Blocks.BLAST_FURNACE, SignalBlockSupplier.container_3()),
+    COMMAND_BLOCK(Blocks.COMMAND_BLOCK, SignalBlockSupplier.command_block()),
+    AUTO(null, null);
+
+    public static SignalBlock getBestBlock(int signal) {
```
Matthias1590, line 22: Could you move this below the constructor?

src/main/java/tools/redstone/redstonetools/utils/SignalBlock.java
```diff
@@ -0,0 +1,39 @@
...
+        if (signal < 1780) return BARREL;
+        else return COMMAND_BLOCK;
```
Matthias1590, line 24: This could just be a ternary statement
```java
return signal <= 1779
    ? BARREL
    : COMMAND_BLOCK;
```

src/main/java/tools/redstone/redstonetools/utils/SignalBlock.java
```diff
@@ -0,0 +1,39 @@
...
+    private final SignalBlockSupplier supplier;
+
+    SignalBlock(Block block, SignalBlockSupplier supplier) {
+        this.block = block;
+        this.supplier = supplier;
+    }
+
+    public ItemStack getItemStack(int signal) {
+        if (block == null || supplier == null) return getBestBlock(signal).getItemStack(signal);
+        else return supplier.get(block, signal);
```
Matthias1590, line 37: the else isnt needed here, could also be a ternary statement but since its quite some code it might be nice to just move the line with else return down 1 and remove the return

src/main/java/tools/redstone/redstonetools/utils/SignalBlock.java
```diff
@@ -0,0 +1,39 @@
...
+    private final Block block;
+    private final SignalBlockSupplier supplier;
+
+    SignalBlock(Block block, SignalBlockSupplier supplier) {
+        this.block = block;
+        this.supplier = supplier;
+    }
+
+    public ItemStack getItemStack(int signal) {
+        if (block == null || supplier == null) return getBestBlock(signal).getItemStack(signal);
```
Matthias1590, line 36: please but the bodies of if statements on the line under the if statement

Matthias1590 (changes requested):

src/main/java/tools/redstone/redstonetools/utils/SignalBlock.java
JohanVonElectrum: SignalBlockSupplier is to keep all the code to generate the nbt together. Each possible NBT has a supplier: command_block, container, composter (And I have to add jukebox in 1.19). What we can do is to remove the `container_x()` suppliers and replace them with calls to `container(x)`.

src/main/java/tools/redstone/redstonetools/features/commands/SignalStrengthBlockFeature.java
```diff
@@ -18,9 +19,7 @@
 
 @AutoService(AbstractFeature.class)
 @Feature(name = "Signal Strength Block", description = "Creates a block with the specified signal strength.", command = "ss")
```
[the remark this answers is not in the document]
JohanVonElectrum: It is fixed in the previous commit: https://github.com/RedstoneTools/redstonetools-mod/pull/271/commits/84bc123451b336c7be5a468b2544b2f699663b87

src/main/java/tools/redstone/redstonetools/features/commands/SsCommandFeature.java
JohanVonElectrum: I think the changes at https://github.com/RedstoneTools/redstonetools-mod/pull/271/commits/aed52ee1c0b2a0e46938e0bf527b71840a9bd42c are sufficient.

JohanVonElectrum:
I need some feedback on this one: https://github.com/RedstoneTools/redstonetools-mod/pull/271#discussion_r1225180354

JayMensink:
Mention the appropriate issue next time.
Issue #246

JohanVonElectrum:
Sorry, I talked about the issue on discord. I didn't know it was open on GitHub. For the next one I will check.

src/main/java/tools/redstone/redstonetools/utils/SignalBlockSupplier.java
```diff
@@ -0,0 +1,85 @@
...
+
+        MutableText text = new LiteralText(String.valueOf(signal));
+        text.setStyle(text.getStyle().withColor(Formatting.RED));
+        item.setCustomName(text);
+        item.addEnchantment(Enchantment.byRawId(0),0);
+
+        return item;
+    }
+
+    static SignalBlockSupplier container(int slots) {
```
Matthias1590, line 35: doesnt adhere to srp, refactoring would probably make it more readable

src/main/java/tools/redstone/redstonetools/utils/SignalBlockSupplier.java
```diff
@@ -0,0 +1,85 @@
...
+import net.minecraft.text.MutableText;
+import net.minecraft.util.Formatting;
+import net.minecraft.util.registry.Registry;
+
+@FunctionalInterface
+public interface SignalBlockSupplier {
+
+    NbtCompound build(int signal);
+
+    default ItemStack get(Block block, int signal) {
```
Matthias1590, line 20: doesnt adhere to srp, which makes it a bit unreadable

src/main/java/tools/redstone/redstonetools/utils/SignalBlockSupplier.java
```diff
@@ -0,0 +1,85 @@
...
+import net.minecraft.item.Items;
+import net.minecraft.nbt.NbtCompound;
+import net.minecraft.nbt.NbtList;
+import net.minecraft.text.LiteralText;
+import net.minecraft.text.MutableText;
+import net.minecraft.util.Formatting;
+import net.minecraft.util.registry.Registry;
+
+@FunctionalInterface
+public interface SignalBlockSupplier {
```
Matthias1590, line 16: why is this an interface with default implementations? i think just a class would be fine

Matthias1590 (changes requested):

src/main/java/tools/redstone/redstonetools/utils/SignalBlockSupplier.java
JohanVonElectrum: I have chosen to use an interface with a build method (could be renamed to getNbt or buildNbt or something similar) and a default get method (could be renamed to getItemStack() or buildItemStack() or something similar) in order to provide a functional interface approach to generate NbtCompund based on an integer signal value, mimicking the Function<Integer, NbtCompound> interface. This allows flexibility and code reuse by allowing different implementations of the SignalBlockSupplier interface.

src/main/java/tools/redstone/redstonetools/utils/SignalBlockSupplier.java
JohanVonElectrum: I understand your concern about SRP and agree that separating responsibilities can improve readability. In the current implementation, the get method is responsible for both creating the ItemStack and applying the NbtCompund. The problem is that I can't think of a better place to put it and I think this is not a bad place at all.

src/main/java/tools/redstone/redstonetools/utils/SignalBlockSupplier.java
JohanVonElectrum: I recognize the need for improved readability and maintainability. The container method and similar static methods serve as factory methods to generate instances of the SignalBlockSupplier interface with predefined behavior. I could move these methods to a class called SignalBlockSuppliers, but I'm not sure if it's worth it just for these 3 methods (although in 1.19 I have to add one more for the jukebox).

src/main/java/tools/redstone/redstonetools/utils/SignalBlockSupplier.java
Matthias1590: still think a class would do just fine

src/main/java/tools/redstone/redstonetools/utils/SignalBlockSupplier.java
Matthias1590: `if (signal < 0) throw new IllegalArgumentException("Signal cannot be negative");` could become `verifySignalStrength(signal)`, `signal` could be renamed to `signalStrength`, `get` should probably be called `getItemStack`, the portion of the code that creates the text could be extracted and put in another function, stuff like that

src/main/java/tools/redstone/redstonetools/utils/SignalBlockSupplier.java
Matthias1590: The location isnt the issue, its just the size of the method, theres constants everywhere that probably dont make a lot of sense to readers, `stack_size` uses snake case for some reason, theres 2 ternary operators in a single line, twice, the for loop is using 4 lines for the `for (...)` part, its just not very readable

Matthias1590 (changes requested):

src/main/java/tools/redstone/redstonetools/utils/SignalBlockSupplier.java
JohanVonElectrum: I get your point on this, I'll see what I get to do with this.

src/main/java/tools/redstone/redstonetools/utils/SignalBlockSupplier.java
JohanVonElectrum: Okay, fair enough. The only problem is with the verifySignalStrength(signal) function. It's complicated because other illegal argument exceptions are thrown elsewhere due to other signal strength validations. Each signal block supplier has its own things to check, like a composter, which can't have signal 7, but can from 0-6 and 8.

src/main/java/tools/redstone/redstonetools/utils/SignalBlockSupplier.java
JohanVonElectrum: I don't understand what you mean exactly.

src/main/java/tools/redstone/redstonetools/utils/SignalBlockSupplier.java
Matthias1590: Then have it take a max ss

src/main/java/tools/redstone/redstonetools/utils/SignalBlockSupplier.java
Matthias1590: we can just use a class instead of an interface

JohanVonElectrum: keeping the interface as stated in discord

src/main/java/tools/redstone/redstonetools/utils/SignalBlockSupplier.java
JohanVonElectrum: The problem with the composter can't be fixed like that, because max 8 included but can't 7. Composters can do [0, 6]U[8]

src/main/java/tools/redstone/redstonetools/utils/SignalBlockSupplier.java
```diff
@@ -0,0 +1,131 @@
...
+        ItemStack item = new ItemStack(block);
+        setCompoundNbt(item, this.createNbt(signalStrength));
+        setItemName(item, signalStrength);
+        return item;
+    }
+
+    static SignalBlockSupplier container(int slots) {
+        return signalStrength -> {
+            if (isInvalidSignalStrength(signalStrength, 1779))
+                throw new IllegalArgumentException("Container signal must be 0-1779");
```
DreiMu, line 31: There are some containers, e.g. furnaces, where this fails in lines 47-48, maybe add a calculation that changes the max value depending on the number of slots

src/main/java/tools/redstone/redstonetools/utils/SignalBlockSupplier.java
```diff
@@ -0,0 +1,131 @@
...
+                    ? 1
+                    : (int) Math.ceil(slots * (signalStrength - 1) / 14D * item.getMaxCount());
+            String itemId = Registry.ITEM.getId(item).toString();
+
+            // Check that the calculated number of items is correct.
+            // This is to prevent problems with items that have a maximum stack size of 1 but stackSize > 1.
+            // TODO: This can be improved by removing an item and adding stackable items up to the desired signal strength.
+            // Even with the improvement, this will still fail for inventories with no available slots.
+            if (calculateComparatorOutput(itemsNeeded, slots, item.getMaxCount()) != signalStrength)
+                throw new IllegalStateException("This signal strength cannot be achieved with the selected container");
```
DreiMu, line 48: It fails here

DreiMu (changes requested):

src/main/java/tools/redstone/redstonetools/utils/SignalBlockSupplier.java
JohanVonElectrum: Intended. The isInvalidSignalStrength method checks the signal range, but containers with less than 15 slots have a discontinuous range, so some values do not work. This is not a bug, it is something impossible to fix, as stated in the comments on line 47.

src/main/java/tools/redstone/redstonetools/utils/SignalBlockSupplier.java
JohanVonElectrum: It's not a bug, it's something on purpose. Read the comments on that line to understand why it happens.

src/main/java/tools/redstone/redstonetools/utils/SignalBlockSupplier.java
JohanVonElectrum: The catch in SignalStrengthBlockFeature for IllegalStateException was missing.

DreiMu (approved):
Until there is a better check/method for filling non-normal containers, this is probably the best solution

PR #277, opened 2023-06-09
Fix QuickTp falling through a block

Nitrrine:
Now we don't fall through block if we looking on solid surface


RUGMJ:
Could you explain this a bit more? How can I recreate the bug so I can test if this fixes it

Nitrrine:
You just need to look on top of the block, that you want to teleport, then after /quicktp you teleporting into block (not on block surface) like, you just falling though it

PR #282, opened 2023-06-11
Signal Strength: Positive Signed Integer Domain

JayMensink:
Issue #246


JayMensink:
Better solution created by Johan.

PR #284, opened 2023-06-11
Internal Feedback Improvements

orbyfied:
No breaking changes to the API.

Feedback now uses and accepts text components for more advanced/customizable formatting. 

You can now send intermediate feedback like warnings with `void Feedback#sendIntermediate` (example: `Feedback.warning("abc {}", abc).sendIntermediate(source)`) and send feedback as the final feedback using `int Feedback#send` when creating commands with Brigadier. The method returns the feedback's status code to be reported to Brigadier. (example: `return Feedback.error("abc {}", abc).send(source)`)

Placeholders are formatted as red by default, but by replacing the default `{}` with `{~}` you can disable the formatting of the placeholder.

I need this for a couple improvements I've been assigned to. At least, it would make it much less painful.

### DIFF
--- a/src/main/java/tools/redstone/redstonetools/features/AbstractFeature.java
+++ b/src/main/java/tools/redstone/redstonetools/features/AbstractFeature.java
@@ -3,6 +3,7 @@ package tools.redstone.redstonetools.features;
 import com.mojang.brigadier.CommandDispatcher;
 import net.fabricmc.fabric.api.command.v1.CommandRegistrationCallback;
 import net.minecraft.server.command.ServerCommandSource;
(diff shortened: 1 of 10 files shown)


Matthias1590:
im sorry but i dont think i can merge this, were gonna refactor the entire project after 2.0.0, also i feel like feedback should only be returned, not directly sent. i think its fine to use this to test the io tui but i dont think we should merge it now if we might end up not using it

orbyfied:
yeah true, i guess these changes should be included into marshal in the future
i do think having the option to send feedback like warnings directly is useful though

also it doesnt even compile appereantly which is probably another git artifact, ill just close this i guess

PR #295, opened 2023-06-17
Add formatter settings for IntelliJ's Auto formatter

varun7654:
The settings are from:
https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml
(Google Java Style)

With the following exceptions:
- Star imports allowed
- Hard wrap increased to 120 characters
- No blank lines after class declaration are required
- Tab size & indent increased to 4 characters
- Continuation Indent increased to 8 characters

(These changes help the auto formatter more closely align with the code already written)

Note: I haven't reformatted any of the already existing code. I can run it if you'd like (as quite a bit of it doesn't follow the guidelines)

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,0 +1,573 @@
+<component name="ProjectCodeStyleConfiguration">
+  <code_scheme name="Project" version="173">
+    <option name="OTHER_INDENT_OPTIONS">
+      <value>
+        <option name="INDENT_SIZE" value="2" />
+        <option name="CONTINUATION_INDENT_SIZE" value="4" />
+        <option name="TAB_SIZE" value="2" />
+      </value>
+    </option>
+    <option name="KEEP_CONTROL_STATEMENT_IN_ONE_LINE" value="false" />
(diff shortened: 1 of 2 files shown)


RUGMJ:
Is there a more general format for multi editor? Eclipse and vscode are both used by other contributors. Vscode supports eclipses format, does intellij support it?

varun7654:
Hi, Sorry for the late reply (I was on vacation).

IntelliJ has a way to export the IntelliJ formatter settings I created into an Eclipse XML, so I think we should be fine if we just have the IntelliJ & Eclipse settings together. 

I don't use Eclipse so I don't know where it creates the files when configuring the autoformatter settings. (If you do, could you enable the autoformatter/change a few settings and take a screenshot of what the file structure it creates looks like so I could put the files in the correct location?)

RUGMJ:
I just checked, eclipse, vscode, intelliJ and neovim all support EditorConfig this would probably be the better choice as its not editor dependent

RUGMJ:
You can also export intellij's format to editor config

varun7654:
> I just checked, eclipse, vscode, intelliJ and neovim all support EditorConfig this would probably be the better choice as its not editor dependent

Ok changed I just changed it to use .editorconfig

PR #298, opened 2023-06-23
Hide macro output

Saturn56:
only works in singleplayer


Saturn56:
Now I added the silent command with the autocomplete and the DisableCommandOutput mixin does the hiding of the output.

PR #313, opened 2023-07-04
add zh-tw.json to "src/main/resources/assets/redstonetools/lang"

LmanTW:
add some translation


Xiang990293:
you can provide translation to RPMTW instead I think.

PR #376, opened 2024-07-02
fix: solve visual bug in air place highlight

coffeeispower:
This is a bug that the 1.20.4 version of this mod had for a long time and was preventing it from actually releasing (at least it seemed to), and i also wanted this bug fixed so I fixed it myself. The issue was that there was a discrepancy of the function that was used between when you actually place a block and when the hightlight is rendered so I made sure both use the same method and the issue is fixed.
imagem


RUGMJ:
lgtm, need someone to test it in game though before it's merged

coffeeispower:
@RUGMJ Well i already sent a screenshot that it is working so like why not just merge?

kr1viah:
Is fixed in 3.0.0, so I'm gonna close this

PR #426, opened 2026-07-31
[PORT] - Add Minecraft 26.2 target

CodeWithCharles:
Adds 26.2 as a seventh Stonecutter target. Also carries three fixes for bugs that already existed on main and got in the way.

**Behaviour change worth reading before merging.** On 26.2, `/ssb` above 15 now gives a command block instead of a container. 26.2 validates container contents strictly, so overstacked slots are silently replaced by EMPTY and you get an empty barrel. Overstacking is how the >15 range worked, and Mojang closed it. Below 26.2 nothing changes.

Commits, oldest first:

- **Run malilib-api-processor on the client source set only.** With `splitEnvironmentSourceSets()` the processor ran twice and wrote `META-INF/kr1v/index.json` at the same jar path both times. Loom <=1.16 silently dropped one, so our shipped 1.21.x jars have only ever contained the client half. Loom 1.17 hard-fails instead, which is how it surfaced. Every malilib-api usage is client side, so scoping the processor there fixes it at the source. Upstream issue: kr1viah/malilib-api#3
- **Build CI on Java 25.** `build_reusable.yml` still installed Java 21 while `publish.yml` was already on 25. The 26.x targets compile with `--release 25`.
- **Fix /ssb command blocks never carrying their signal.** SuccessCount was written to `CUSTOM_DATA`, which is item level data. `BLOCK_ENTITY_DATA` is what gets copied into the block entity on placement, so every /ssb command block has been outputting 0. Affects all seven versions, fixed on all seven.
- **Bump Gradle to 9.5.1 and Loom to 1.17.** Required by 26.2 per the Fabric announcement.
- **Add Minecraft 26.2 target.** Version file plus the API changes: dyed shulker box items collapsed into a ColorCollection, screen handling moved off Minecraft onto Gui, LevelRenderer.renderHitOutline replaced by the submit node model, and GameRenderer.getMainCamera renamed.

Tested in game on 26.2, 26.1.2 and 1.21.4 (ssb across all versions): mod loads, airplace outline renders in the right place, config menu and macro editor, macros run, //rstack, //colorcode, /ssb across the range.

One thing that is not new but was never written down anywhere: a command block only keeps its SuccessCount if you place it in creative with permission level 2. Vanilla restriction on command block data, applies on every version.

> [!IMPORTANT]
Here's an attachement that contains the diff file to apply to the wiki :
[diff.txt](https://github.com/user-attachments/files/30593578/diff.txt)

### DIFF
--- a/src/main/resources/fabric.mod.json
+++ b/src/main/resources/fabric.mod.json
@@ -37,7 +37,7 @@
   "depends": {
     "fabricloader": ">=0.18.1",
     "fabric-api": "*",
-    "minecraft": ">=1.21.4 <=${minecraft_version_out}"
+    "minecraft": ">=1.21.4 <=26.2"
   },
   "recommends": {
     "worldedit": "*",
--- a/src/main/resources/fabric.mod.json
+++ b/src/main/resources/fabric.mod.json
@@ -37,7 +37,7 @@
   "depends": {
     "fabricloader": ">=0.18.1",
     "fabric-api": "*",
-    "minecraft": ">=1.21.4 <=26.1.2"
+    "minecraft": ">=1.21.4 <=${minecraft_version_out}"
   },
   "recommends": {
     "worldedit": "*",
--- a/.github/workflows/build_reusable.yml
+++ b/.github/workflows/build_reusable.yml
@@ -22,7 +22,7 @@ jobs:
             - name: Setup JDK
               uses: actions/setup-java@v4
               with:
-                  java-version: 21
+                  java-version: 25
                   distribution: 'microsoft'
 
             - name: Setup Gradle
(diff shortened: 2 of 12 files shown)


src/main/resources/fabric.mod.json
```diff
@@ -37,7 +37,7 @@
   "depends": {
     "fabricloader": ">=0.18.1",
     "fabric-api": "*",
-    "minecraft": ">=1.21.4 <=26.1.2"
+    "minecraft": ">=1.21.4 <=${minecraft_version_out}"
```
kr1viah: This should stay/become 26.2

kr1viah (changes requested):

src/main/resources/fabric.mod.json
CodeWithCharles: Yeah you're right. The template also narrows it on every other jar, 1.21.4 would go from <=26.1.2 to <=1.21.4 and stop loading on 1.21.11, which changes behaviour for versions this PR shouldn't touch. Hardcoding 26.2.

CodeWithCharles: Should be good !

kr1viah (approved):

